PR #6, opened 2019-10-21
update test/expected_constants_n25_s0.txt

jgphpc:
Could you confirm if `test/expected_constants_n25_s0.txt` needs to be updated?

I am running the develop branch:

```
./gnu/omp.app -n 25 -s 0
cat constants.txt

0 1.1e-06 1.1e-06 2.0801e+10 2.08e+10 1e+06 6636775
```

do you get the same ?


acavelan:
I get test FAILED, but values are the same. I am not sure what is wrong with the test, perhaps an extra space?

```
Correctness test FAILED
Output file diff:
1c1
< 0 1.1e-06 1.1e-06 2.0801e+10 2.08e+10 1e+06 6637375 
---
> 0 1.1e-06 1.1e-06 2.0801e+10 2.08e+10 1e+06 6636775 

```

Anyway, I just checked with the newTree branch and I still get the the same values, so yes.

jgphpc:
The test fails because the number in the last column differ:

```
expected: 6637375 
got: 6636775
```

acavelan:
Ah! Indeed. It seems that constants.txt needs to be updated.

acavelan:
We will need to update the value and the test again.

PR #12, opened 2019-11-13
Global tree creation

acavelan:
The consequence of this update is a reduced number of halos particles for the first iteration. This reduces the amount of memory needed (in the first iteration), and makes it much faster too.

Using size_t everywhere we deal with large number of particles enable the mini-app to handle large test cases. It has been tested on Piz Daint with 32,000,000,000 particles on 1024 nodes (12 cores + GPU).


src/include/DistributedDomain.hpp
```diff
@@ -63,19 +64,19 @@ class DistributedDomain
         }
     }
 
-    int64_t neighborsSumImpl(const Task &t)
+    size_t  neighborsSumImpl(const Task &t)
     {
-        int64_t sum = 0;
+        size_t  sum = 0;
 #pragma omp parallel for reduction(+ : sum)
         for (unsigned int i = 0; i < t.clist.size(); i++)
```
michalgrabarczyk: Should be size_t instead of unsigned int

src/include/DistributedDomain.hpp
```diff
@@ -469,48 +467,68 @@ class DistributedDomain
...
+
+        // build the global tree using only the particles that belong to us
+        octree.globalRebalance(d.bbox.xmin, d.bbox.xmax, d.bbox.ymin, d.bbox.ymax, d.bbox.zmin, d.bbox.zmax);
         octree.buildGlobalTreeAndGlobalCountAndGlobalMaxH(clist, x, y, z, h, ordering);
+        octree.assignProcesses(work, work_remaining);
+
+        // Get rid of particles that do not belong to us
         reorder(ordering, d);
+        for (size_t i = 0; i < workAssigned; i++)
+            clist[i] = i;
```
michalgrabarczyk: std::iota could be used here

src/include/DistributedDomain.hpp
```diff
@@ -522,49 +540,29 @@ class DistributedDomain
...
+        const size_t remaining = ntot - comm_size * split;
 
         d.bbox.computeGlobal(clist, d.x, d.y, d.z);
 
-        std::vector<int> work(comm_size, split);
+        std::vector<size_t> work(comm_size, split);
         work[0] += remaining;
 
         // We map the nodes to a 1D array and retrieve the order of the particles in the tree
         std::vector<int> ordering(n);
```
michalgrabarczyk: Shouldn't it be size_t?

src/include/utils.hpp
```diff
@@ -1,29 +1,29 @@
-#pragma once
+// #pragma once
```
michalgrabarczyk: This file should be deleted if not used

PR #14, opened 2019-12-05
Fixes

acavelan:
- fix major bug with PBC 'on' in positions.hpp
- fix bug in Printer with dt[0] when node has 0 particles
- remove empty utils.hpp file
- add back variable smoothing length (thanks to fixing major bug 1.)
- add maxvsignal for timestep calculation (needed for evrard and wind bubble)


src/include/DistributedDomain.hpp
```diff
@@ -685,20 +685,21 @@ class DistributedDomain
...
 
-#pragma omp parallel for
-        for (int pi = 0; pi < n; pi++)
+        size_t  n = t.clist.size();
+
+#pragma omp parallel for schedule(guided)
+        for (size_t pi = 0; pi < n; pi++)
         {
-            const int i = clist[pi];
+            int i = t.clist[pi];
```
michalgrabarczyk: Some things could be const here. The i, n and I think also the Task &t function parameter. If Task &t can be const, then task variable in for loop in line 716 can be const as well

src/include/ParticlesData.hpp
```diff
@@ -52,15 +53,11 @@ struct ParticlesData
     constexpr static T Kcour = 0.2;
     constexpr static T maxDtIncrease = 1.1;
     const static T K;
-    static T dx;
 
     // settings
     constexpr static ushort noOfGpuLoopSplits = 4; // No. of loop splits running in GPU to fit into the GPU memory
```
michalgrabarczyk: That's not used anymore I think and can also be deleted

PR #15, opened 2019-12-05
Domain separation

acavelan:
- split DistributedDomain into Domain and DistributedDomain
- add findNeighbors module (previously in Domain)
- add updateSmoothinglength module (previously in Domain)


src/include/sph/findNeighbors.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <vector>
+#include "Task.hpp"
+
+namespace sphexa
+{
+namespace sph
+{
+
+template <typename T, class Dataset>
+void findNeighborsImpl(Task &t, Dataset &d, const Octree<T> &o)
```
michalgrabarczyk: Just for the sake of consistency, output parameters should be the last arguments

src/include/sph/findNeighbors.hpp
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+#ifdef USE_MPI
+    MPI_Allreduce(MPI_IN_PLACE, &sum, 1, MPI_LONG_LONG_INT, MPI_SUM, MPI_COMM_WORLD);
+#endif
+
+    return sum;
+}
+} // namespace sph
+} // namespace sphexa
```
michalgrabarczyk: There should be extra new line on the end of each file

src/sqpatch.cpp
```diff
@@ -67,20 +69,19 @@ int main(int argc, char **argv)
         timer.step("UpdateQuantities");
         sph::computeTotalEnergy<Real>(taskList, d);
         timer.step("EnergyConservation"); // AllReduce(sum:ecin,ein)
-        distributedDomain.updateSmoothingLength(taskList, d);
+        sph::updateSmoothingLength<Real>(taskList, d);
         timer.step("UpdateSmoothingLength");
 
-        long long int totalNeighbors = distributedDomain.neighborsSum(taskList);
+        size_t totalNeighbors = sph::neighborsSum(taskList);
```
michalgrabarczyk: Could be const

PR #42, opened 2020-09-23
Gpu hackathlon no copy

cypox:
computing neighbors in gpu without sending data from host


sebkelle1:
Changes look good. Should I wait with until the cudaAllocs/Frees are grouped in the TaskList?

sebkelle1 (approved):

PR #49, opened 2020-09-28
Weak scaling

acavelan:
This branch is based on the gpu-hack-noallocs branch.

- Reduce memory footprint of the tree (no longer the bottleneck)
- Remove unnecessary data (ro_0 and p_0)
- Remove initFluidDensityAtRest call
- Fix work allocation and load balancing for both small and high number of ranks
- Fix a debug macros and add some more
- Fix Global Tree configuration for running at scale


include/ParticlesData.hpp
```diff
@@ -63,6 +124,141 @@ struct ParticlesData
...
+#endif
+
+
+#if defined(USE_CUDA)
+    // number of CUDA streams to use
+    static const int NST = 2;
+
+    cudaStream_t streams[NST];
+    
+    int *d_clist[NST], *d_neighbors[NST], *d_neighborsCount[NST]; // work arrays per stream
```
sebkelle1, line 180: the cuda streams and the d_clist, d_neighbors and d_neighborsCount could go into a separate
struct `Workspace`. Then you can do `Workspace streamScratch[NST]`. Each workspace takes
care of allocation and deallocation.

include/ParticlesData.hpp
```diff
@@ -63,6 +124,141 @@ struct ParticlesData
...
+    T *d_x, *d_y, *d_z, *d_vx, *d_vy, *d_vz, *d_m, *d_h, *d_ro, *d_p, *d_c, *d_c11, *d_c12, *d_c13, *d_c22, *d_c23, *d_c33, *d_wh, *d_whd;
+    BBox<T> *d_bbox;
+    T *d_grad_P_x, *d_grad_P_y, *d_grad_P_z, *d_du, *d_maxvsignal;
```
sebkelle1, line 183: From a software design point of view, all this device code should go into its own `struct`, that takes care of allocation/deallocation with RAII (https://en.cppreference.com/w/cpp/language/raii)
because that achieves better separation of concerns.

sebkelle1: the `moveParticleDataToDevice` will still be a member of this new struct and receive the `ParticleData` with the CPU host data as an argument.

PR #51, opened 2020-10-01
fix cpu density bug

sebkelle1:
This must be a bug...


acavelan:
Yep, looks like a bug to me. The ro[i] is the correct one.

acavelan (approved):

Issue #54, opened 2020-10-01
MPI IO missing particle with more than 1 MPI rank


acavelan:
In MPIFileutils.hpp, the code has a wrong assumption on the number of particles per node.

```
const size_t split = d.n / d.nrank;
const size_t remaining = d.n - d.nrank * split;

const MPI_Offset col = d.n * sizeof(double);
MPI_Offset offset = d.rank * split * sizeof(double);
if (d.rank == 0) offset += remaining * sizeof(double);
```

jgphpc:
@acavelan can't reproduce, how many mpi ?

acavelan:
2 MPI nodes are enough. You can use the script/plot.py to plot the data. The file has the right size, but you will see that (most of the time, but not always) particles are missing (nothing is displayed in some corner of the space).

jgphpc:
see https://github.com/unibas-dmi-hpc/SPH-EXA_mini-app/issues/60 too

PR #58, opened 2020-10-16
Add hdf5 support

jgphpc:
Fix for #54 (does not fix mpi i/o, just adding parallel-hdf5 that works)


cypox:
I will merge with the new develop

PR #59, opened 2020-10-20
GPU with Ressource Acquisition Is Initialization

cypox:
Restructured the code to initialize GPU memory only at start and free the memory at the very end.

I tested this branch on single and 2 nodes. The execution time should also be a smidge faster.

gpu-raii:
32.25
19.86
20.95
20.42

gpu-hackathlon:
33.46
20.50
21.05
20.52


include/DistributedDomain.hpp
```diff
@@ -424,6 +424,11 @@ class DistributedDomain : public Domain<T, Dataset, Tree>
         // MUST DO SYNCHALOS AND BUILDTREE AFTER
 
         d.count = workAssigned;
+
+#if defined(USE_CUDA)
```
jgphpc, line 428: why not ifdef ?

include/Domain.hpp
```diff
@@ -7,6 +7,10 @@
 #include "Octree.hpp"
 #include "Task.hpp"
 
+#if defined(USE_CUDA)
```
jgphpc, line 10: why not ifdef ?

jgphpc:
Can you add:
```
#ifdef USE_CUDA
#include "cuda/sph.cuh"
#endif
```
in: `include/sph/findNeighbors.hpp` and `include/sph/density.hpp` ?

include/DistributedDomain.hpp
cypox: Just in case we needed an other condition. With ifdef you can only check for one flag. If we need to check for USE_CUDA and USE_MPI it'll be cumbersome using ifdefs.

jgphpc: ok, then I would update the other ifdef to have the same everywhere.

PR #65, opened 2021-01-25
shared densityJLoop between CPU and GPU

sekelle:


cypox (approved):
That's a better solution.

I will merge these changes and do the same for the remaining kernels.

PR #72, opened 2021-03-09
Fix non-cuda compilation with CMake, reoganize domain file names

sekelle:
-fixes the compilation of the sedov test case without CUDA
-removes dependency on `std::exclusive_scan` as not all compilers support it yet
-cleaner file/directory layout for the cornerstone domain. Instead of a single include directory for all the code,
there are now domain/, halos/, tree/, primitives/ and sfc/ subdirectories.


jgphpc:
This is going to be my reference version of the code for ci/cd tests.
Non gpu test is already looking good:

```
[       OK ] (1/4) SphExa_timers_Check_0_1_10000_0 on pilatus:mc using cpeGNU
[       OK ] (2/4) SphExa_timers_Check_0_1_10000_0 on pilatus:mc using cpeAMD
[       OK ] (3/4) SphExa_timers_Check_0_1_10000_0 on pilatus:mc using cpeCray
[       OK ] (4/4) SphExa_timers_Check_0_1_10000_0 on pilatus:mc using cpeIntel
```

cypox:
Checked!

cypox (approved):
Checked

PR #83, opened 2021-06-18
Clang formatting

sekelle:


jgphpc (approved):
tested with cdt/21.05 (PrgEnv-nvidia, PrgEnv-gnu)

PR #87, opened 2021-07-16
Flexible exchange

sekelle:
This PR introduces a more flexible domain-exchange algorithm where new particles after a domain decomposition
update are written into buffer areas that were used for halo particles in the previous time step.
This allows to defer the reordering of particles into SFC-order until after halo discovery, such that the reordering
can directly put the particles into their final locations.


jgphpc (approved):
Unit tests passed :ok: on dom with [gitlab](https://git.cscs.ch/piccinal/SPH-EXA_mini-app/-/jobs/127630#L28)

```
gcc/9.3.0
cudatoolkit/11.2.0_3.39-2.1__gf93aa1c
cdt-cuda/21.05
```

PR #88, opened 2021-07-22
Improve compatibility with the new NVHPC SDK

sekelle:
Since the project enables CUDA as a language, we don't need to also find the CUDAToolkit
in order to be able to link executables against -lcudart.
This fixes compilation with nvhpc on Ault.


jgphpc (approved):
gitlab job=127944 succeeded on dom (PrgEnv-gnu)

PR #89, opened 2021-08-03
Add cmake install targets

jgphpc:


jgphpc:
TODO: cuda targets too

PR #105, opened 2021-10-15
added unit test for density kernel

sekelle:


jgphpc (changes requested):
temporary blocking merge to give me some time to validate/debug the ci (I will unlock when ready).

include/sph/test/kernel/density.cpp
```diff
@@ -0,0 +1,130 @@
...
+    std::vector<int> neighborsCount{4};
+
+    std::vector<T> x{1.0, 2.1, 3.2, 4.3, 5.4};
+    std::vector<T> y{1.1, 2.2, 3.3, 4.4, 5.5};
+    std::vector<T> z{1.2, 2.3, 3.4, 4.5, 5.6};
+    std::vector<T> h{5.0, 5.1, 5.2, 5.3, 5.4};
+    std::vector<T> m{1.1, 1.2, 1.3, 1.4, 1.5};
+
+    /* distances of particle zero to particle j
+     *
```
rmcabezon, line 66: With these distances and h(0)=5, all particles will be neighbors. I'd recommend using a smaller h so that some particles are neighbors and others are not. In that way, you ensure also that the kernel is adding a zero when a particle is qualified as a neighbor but in fact, is not. This might happen if we define the neighborhood with a gather models instead of a scatter model.

include/sph/test/kernel/density.cpp
```diff
@@ -0,0 +1,130 @@
...
+    std::vector<int> neighbors{1, 2, 3, 4};
+    std::vector<int> neighborsCount{4};
+
+    std::vector<T> x{1.0, 1.1, 1.4, 9.9, 10.4};
+    std::vector<T> y{1.1, 1.2, 1.5, 9.8, 10.2};
+    std::vector<T> z{1.2, 1.3, 1.6, 9.7, 10.3};
+    std::vector<T> h{2.5, 2.51, 2.52, 2.53, 2.54};
+    std::vector<T> m{1.1, 1.2, 1.3, 1.4, 1.5};
+
+    /* distances of particle 0 to particle j
```
rmcabezon, line 111: Same comment as above. Either reduce h(0) or add some particles farther than 2h(0).

rmcabezon (changes requested):

include/sph/test/kernel/density.cpp
sekelle: The current implementation of the density kernel assumes that the neighbor list is correct. Whatever is listed as a neighbor will be added, even if the distance turns out to be bigger than 2*h. So I can't really test correct neighbor search here.

sekelle:
@jgphpc This MR need not be held up just because the test is not in the CI yet. You can add it to the CI in a separate MR.

jgphpc (approved):
[  PASSED  ] Ran 7/8 test case(s) from 8 check(s) (0 failure(s), 1 skipped)
Will add new unittest after merging.

rmcabezon (approved):

jgphpc:
Check gitlab tests results before merging: [  PASSED  ] Ran 31/32 test case(s) from 32 check(s) (0 failure(s), 1 skipped) https://git.scicore.unibas.ch/j.piccinali/mirror_test/-/pipelines/13401 github_pr=105
fa8c245

PR #107, opened 2021-10-18
Cmake fix

biddisco:
A couple of minor cmake fixes to get it building locally

### DIFF
--- a/src/evrard/CMakeLists.txt
+++ b/src/evrard/CMakeLists.txt
@@ -1,10 +1,16 @@
+# this target uses cuda but isn't compiled with cuda compiler
+# so add the right cuda runtime lib
+find_library(CUDART_LIBRARY cudart ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES})
 
 add_executable(evrard evrard.cpp)
 target_include_directories(evrard PRIVATE ${PROJECT_SOURCE_DIR}/domain/include)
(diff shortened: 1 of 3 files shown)


jgphpc:
```
[  PASSED  ] Ran 33/34 test case(s) from 34 check(s) (0 failure(s), 1 skipped)
```

jgphpc (approved):

src/evrard/CMakeLists.txt
```diff
@@ -1,10 +1,16 @@
+# this target uses cuda but isn't compiled with cuda compiler
+# so add the right cuda runtime lib
+find_library(CUDART_LIBRARY cudart ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES})
```
sekelle, line 3: Let's add `find_package(CUDAToolkit)` just after we enable the CUDA language. Then we can
link applications against `CUDA::cudart` and omit adding the include dirs explicitly.

sekelle (changes requested):

src/evrard/CMakeLists.txt
biddisco: I actually tried that, but for reasons I didn't want to spend (waste) time trying to understand, `find_package(CUDAToolkit)` didn't work on my laptop, so I left it as find library for now. I might have made a silly typo or something and should try again really.

sekelle:
superseded by #109

PR #115, opened 2021-10-29
Neighbor opt cuda

sekelle:
This fuses the GPU SPH and neighbor search kernels, removes the need for pre-allocated neighbor lists
and replaces the local particles list with a particle range.


jgphpc:
Check gitlab tests results before merging: [  FAILED  ] Ran 1/46 test case(s) from 46 check(s) (1 failure(s), 0 skipped) https://git.scicore.unibas.ch/j.piccinali/mirror_test/-/pipelines/13489 github_pr=115
67f5532

PR #126, opened 2021-11-19
Artificial viscosity update

sekelle:
Uses Sphnx's implementation of the artificial viscosity. Fixes noisy post-shock radial profiles in the Evrard collapse.

### DIFF
--- a/include/sph/kernels.hpp
+++ b/include/sph/kernels.hpp
@@ -59,7 +59,6 @@ CUDA_DEVICE_FUN inline T artificial_viscosity_old(T ro_i, T ro_j, T h_i, T h_j, 
     T viscosity_ij = 0.0;
     if (rv < 0.0)
     {
-        // calculate muij
         T mu_ij = (h_ij * rv) / (r_square + epsilon * h_ij * h_ij);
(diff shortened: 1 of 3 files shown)


include/sph/kernels.hpp
```diff
@@ -51,16 +56,39 @@ CUDA_DEVICE_FUN inline T artificial_viscosity(T ro_i, T ro_j, T h_i, T h_j, T c_
...
+ * @tparam T    float or double
+ * @param c_i   speed of sound particle i
+ * @param c_j   speed of sound particle j
+ * @param w_ij  relative velocity (v_i - v_j), projected onto the connecting axis (r_i - r_j)
+ * @return      the viscosity
+ */
+template<typename T>
+CUDA_DEVICE_FUN inline T artificial_viscosity(T c_i, T c_j, T w_ij)
+{
+    // these are const now, but will be different for each particle when using viscosity switching
```
rmcabezon: Maybe state here that only alfa will be variable. Beta will always be 2.

rmcabezon (approved):

include/sph/kernels.hpp
sekelle: ok, will do.

PR #127, opened 2021-11-22
simpler gitlab-ci

jgphpc:


jgphpc:
current setup does not allow pr from forks

PR #134, opened 2021-12-16
Update cmake for HIP (sedov)

jgphpc:
Tested on Mi100 with rocm-4.3.1 (requires that the src files be converted with hipify-perl first).


CMakeLists.txt
```diff
@@ -31,6 +31,31 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
 endif()
```
sekelle, line 34: I would add an option here, e.g.
```suggestion
option(SPH_EXA_WITH_HIP "Build SPH-EXA with HIP" OFF)
```
and then check for / enable HIP if that option is ON.
Otherwise you'll not be able to compile with CUDA on a system that has both CUDA and HIP installed as HIP
would always be chosen.

sekelle (changes requested):

CMakeLists.txt
jgphpc: https://github.com/unibas-dmi-hpc/SPH-EXA/pull/134/commits/a233cd64b210cbef0120e6d479aa629f9c9bef6d

sekelle (approved):

PR #135, opened 2021-12-17
Improve sedov and evrard test runs

jaescartin1:
- Sedov cosmetic changes
- Evrard some fixes to run. Add the initial input file (Test3DEvrardRel.bin)
- Remove some warnings


jgphpc:
remove this Test3DEvrardRel.bin please

src/evrard/evrard_new.cpp
```diff
@@ -42,13 +47,13 @@ int main(int argc, char** argv)
...
 
-    const size_t maxStep = parser.getInt("-s", 10);
     const size_t nParticles = parser.getInt("-n", 65536);
+    const size_t maxStep = parser.getInt("-s", 10);
     const int writeFrequency = parser.getInt("-w", -1);
     const int checkpointFrequency = parser.getInt("-c", -1);
     const bool quiet = parser.exists("--quiet");
     const std::string checkpointInput = parser.getString("--cinput");
-    const std::string inputFilePath = parser.getString("--input", "bigfiles/Test3DEvrardRel.bin");
+    const std::string inputFilePath = parser.getString("--input", "../../../src/evrard/bigfiles/Test3DEvrardRel.bin");
```
sekelle: we could change this to
```suggestion
    const std::string inputFilePath = parser.getString("--input", "./Test3DEvrardRel.bin");
```

sekelle (changes requested):

sekelle:
Please remove the large binary file from the PR. The rest looks fine.

jaescartin1:
I removed the initial file './Test3DEvrardRel.bin'. Although I put this file as input, the evrard execution shows some errors. 
Is it possible to execute without this initial file? 
Please, How do you do?

domain/test/unit/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ set(UNIT_TESTS
         test_main.cpp)
 
 add_executable(component_units ${UNIT_TESTS})
-target_compile_options(component_units PRIVATE -Wall -Wextra)
+target_compile_options(component_units PRIVATE -Wall -Wextra -Wno-unknown-pragmas)
```
sekelle: No longer needed, this has just been merged into develop.

sekelle:
It's ok to rename the old evrard.cpp file, but the renaming should be done in a follow-up PR. Because if you both introduce changes in evrard_new.cpp and rename the file, the diff will be obscured.
So in this PR, you can delete evrard.cpp, but do the renaming of evrard_new to evrard in a follow up.
This will also make it easier for @jgphpc to review the CI changes.

sekelle:
> I removed the initial file './Test3DEvrardRel.bin'. Although I put this file as input, the evrard execution shows some errors. Is it possible to execute without this initial file? Please, How do you do?

The file is needed as input. But you could also provide any other file with valid initial conditions, with any number of particles.
Note that running with more than 1 rank is not enabled yet for this test case.

I'll update the README.md file to describe how the binary file can be generated from the .txt with 3 lines of python.

jaescartin1:
Perfect, thanks. I need to provide the analytical solution for evrard test and I need to know how to execute properly the test in the code.

I just see that is difficult to read the changes removing and renaming the old cpp of the evrard test. In the review code, the file is mixed with the previous one ... and I understood that I will be difficult to verify. As you suggested, next time I will do this kind of changes in two steps.

jgphpc (approved):

sekelle (approved):

PR #140, opened 2021-12-25
Analytical sedov

jaescartin1:
Added analytical solution for the Sedov test.

It generate a file with the solution in 1D for the next geometries: planar(1D), cylindrical(2D) or spherical(3D)


sekelle:
Hi Jose, thanks for your work. I'll be excited to see the comparison with an analytical reference.
I think the analytical solution should be an independent app, not coupled to `sedov.cpp` . You could then have a python or bash script that combines the data from the sedov app and the analytical reference and compare them.

jaescartin1:
Hi Sebastian, the comparative is not completely ready yet. The solution works with a planar (1D), cylindrical (2D) or spherical (3D) symmetry.  It means that we still need to modify the sedov sph-exa input to provide a spherical symmetry 3D model. This is the previous step to work with Ruben an reach the final solution.

jaescartin1:
I attached the original fortran code , fully tested in the folder 'src/sedov/analytical', we should remove this folder in the future.
The 'SedovAnalyticalSolution.hpp' file is my c++ adaptation from the original code.

jaescartin1:
Ruben asked me to integrate the Sedov solution inside the sph-exa code, in order to generate automatically the solution when the user select dump and provide the solution flag. It is much more easy and useful to compare the simulate and the analytical output because the solution use automatically the same input parameters than sph-exa on that iteration and time.

jaescartin1:
If you think that everything seems correct, and you don't mind, please approve the pull request and we can modify the implementation in the future if it is need. I would like to have it in the develop branch for the moment to test it the code with Ruben.

jaescartin1:
Hi Sebastian, the Sedov exact solution is in place and working. We can merge the branch.

jgphpc:
cscs-ci run

jgphpc:
Some tests failed the CI: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/SPH-EXA-44692846847247/-/pipelines/452889762

jaescartin1:
Hi Jean-Guillaume,
      thanks for check it.  I just removed a function to reduce the size to a sphere. It seems didn't match well with the GPU implementation after the reduccion. Like this function is not needed for the test, I just removed it.

I hope now the modifications pass the tests.

jgphpc:
cscs-ci run

sebkelle1:
Hi @jaescartin1, thanks for your contribution.
The analytical solution itself and the changes to the SedovDataGenerator are fine.

One problem that remains is that the code for the analytical solution is introduced into the main compilation unit of sedov.cpp.
We have to avoid this coupling, because what's now just the sedov test case in sedov.cpp will evolve into a more general simulation driver that will be able to handle any hydro-only type of simulation, including the turbulence tests on LUMI.
We don't actually want to repeat the whole time-stepping loop for every new set of initial conditions. Therefore we can't have
code for the analytical solution introduced in sedov.cpp

The sedov analytical solution can and should be a stand-alone application: You first execute a run with the sedov application, writing snapshots at the times where you want to compare with the analytical solutions.
Next, `sedov-analytical` can be executed. It includes the same sedov_data_generator.hpp file as sedov.cpp to get the simulation parameters and it reads the required simulation data (x,y,z,vx,vy,vz,time) from the output that the sedov solution has generated.

sekelle (changes requested):
Hi @jaescartin1, thanks for your contribution.
The analytical solution itself and the changes to the SedovDataGenerator are fine.

One problem that remains is that the code for the analytical solution is introduced into the main compilation unit of sedov.cpp.
We have to avoid this coupling, because what's now just the sedov test case in sedov.cpp will evolve into a more general simulation driver that will be able to handle any hydro-only type of simulation, including the turbulence tests on LUMI.
We don't actually want to repeat the whole time-stepping loop for every new set of initial conditions. Therefore we can't have
code for the analytical solution introduced in sedov.cpp

The sedov analytical solution can and should be a stand-alone application: You first execute a run with the sedov application, writing snapshots at the times where you want to compare with the analytical solutions.
Next, sedov-analytical can be executed. It includes the same sedov_data_generator.hpp file as sedov.cpp to get the simulation parameters and it reads the required simulation data (x,y,z,vx,vy,vz,time) from the output that the sedov solution has generated.

sekelle:
Note, I would set up `sedov-analytical` such that it compares only a single time/snapshot and control comparing multiple times with a python script.

src/sedov/SedovDataGenerator.hpp
```diff
@@ -12,6 +12,16 @@ template <typename T, typename I>
...
 {
 public:
+
+    static constexpr double gamma         = 5.0/3.0;
+    static constexpr double r0            = 0.;
+    static constexpr double r1            = 0.5;
+    static constexpr double energytot     = 1.0;
+    static constexpr double width         = 0.10;
+    static constexpr double rho0          = 1.0;
+    static constexpr double ener0         = energytot / std::pow(M_PI,(3.0/2.0)) / 1.0 / std::pow(width,3);
```
sekelle: How is this not generating a compiler error in the CI? This should not compile because `std::pow` is not constexpr,
and in fact does not compile for me.

jaescartin1: I couldn't see this problem with make or cmake in my Ubuntu.
Do you use some additional flags for the compilation?

Maybe is a different compiler version. This is mine:
$ mpic++ --version
g++ (Ubuntu 11.2.0-7ubuntu2) 11.2.0

jaescartin1: As far as I can see on my system, the pow(x,y) function is taken from the include <math.h> at '/usr/x86_64-linux-gnu/include/bits/mathcalls.h'

$ cat /usr/x86_64-linux-gnu/include/bits/mathcalls.h
...
/* Return X to the Y power.  */
__MATHCALL_VEC (pow,, (_Mdouble_ __x, _Mdouble_ __y));
...

sekelle: It works with GCC and fails with clang. According to the C++17 standard, std::pow is not constexpr, so GCC is wrong and clang is right. This should be:
`static inline const double energ0 = ...`

sekelle: @jgphpc are we only testing with GCC?

jgphpc: yes but I can add clang

jgphpc:
cscs-ci run

jaescartin1:
The required changes are in-place. Please, check it again.

jgphpc:
cscs-ci run

src/analyticalSolutions/compare_solutions.py
```diff
@@ -0,0 +1,304 @@
...
+        plt.xlabel('r')
+        plt.ylabel('rho')
+        plt.draw()
+        plt.title('Density')
+        plt.legend(loc='upper right')
+        plt.savefig(figureName, format='svg')
+        plt.figure().clear()
+        print("'Radius vs Density' done.")
+    
+        figureName = "./sedov_presure_comparation_"   + time.__str__() + ".svg"
```
sekelle: perhaps `sedov_pressure` is enough (note double ss) ?

sekelle:
Some observations:

- An SVG file with 125000 (and more) particles is computationally very demanding to render, we should perhaps make .png the default
- File naming convention: the most widely used naming scheme is `lower_case_with_underscore`, e.g. `sedov_solution.cpp`, I already applied this convention everywhere in `domain/`, but not yet in `include/`
- I invoked the python script with `./compare_solutions.py sedov --binary_file ./sedovSolution --nparts 125000 --snapshot_file ./dump_sedov200.txt --constants_file ./constants.txt`, i.e. I forgot to also supply `--iteration 200`. In this case the script assumed iteration 0 as the default, so I got a wrong plot. Perhaps we should make `iteration` a mandatory argument if `constants_file` is given

sekelle:
When building with cmake, the `compare_solutions.py` script is not copied to the installation directory, so I had to copy that by hand. Perhaps @jgphpc could advise how to modify CMakeLists.txt to also install the python script?

src/analyticalSolutions/compare_solutions.py
```diff
@@ -0,0 +1,304 @@
...
+        file.close()
+        
+    
+
+    # Plot graphics
+    if (not no_plots):
+        
+        print("\nGenerating graphics ...")
+        
+        figureName = "./sedov_density_comparation_"   + time.__str__() + ".svg"
```
jgphpc: comparison

src/analyticalSolutions/sedovSolution/main.cpp
```diff
@@ -0,0 +1,181 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+/*! @file
+ *
+ * @brief This program generate the analytical sedov solution based in the time and the initial conditions
+ *
+ * @author Jose A. Escartin <ja.escartin@gmail.com>
+ *
```
jgphpc: You may want to add a reference:
https://cococubed.com/papers/kamm_2000.pdf

jaescartin1: Hi, yes sure, but this is already added but not in the 'main.ccp' where only is called. The reference is in the solution header itself 'SedovSolution.hpp'

 *        This routine is a C++ conversion of one Fortran code based in these two papers:
 *        - "evaluation of the sedov-von neumann-taylor blast wave solution", Jim Kamm, la-ur-00-6055
 *        - "the sedov self-similiar point blast solutions in nonuniform media", David Book, shock waves, 4, 1, 1994

jaescartin1: The Frank Timmes code (in the source that you pointed out) is a evolution, wrote in 2019, of the base code that I used.
https://cococubed.com/research_pages/sedov.shtml

jaescartin1: I ported and removed the bad practices (as goto jumps!) but maybe need a more deep refactoring to be more nice and readable.

jaescartin1: Checking the results in 1D, 2D and 3D, both codes generate the same outputs but with a small diferences done by the numerical noise introduced for the conversion from Fortran to C++

jaescartin1: Checking the results in 1D, 2D and 3D, both codes generate the same outputs but with a small differences done by the numerical noise introduced for the conversion from Fortran to C++

jgphpc:
> the `compare_solutions.py` script is not copied to the installation directory

see https://github.com/unibas-dmi-hpc/SPH-EXA/pull/140/commits/c58d9ecadf51dead99474f743fe2cd132f46a835

jgphpc:
cscs-ci run

src/analyticalSolutions/compare_solutions.py
sekelle: also here, perhaps we can just use `sedov_density_` for brevity.

jgphpc:
cscs-ci run

src/analyticalSolutions/sedovSolution/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+set(SOURCES FileData.cpp SedovSolution.cpp main.cpp)
+add_executable(sedovSolution ${SOURCES})
+target_include_directories(sedovSolution PRIVATE ${PROJECT_SOURCE_DIR}/include)
+target_include_directories(sedovSolution PRIVATE ${PROJECT_SOURCE_DIR}/src)
+install(TARGETS sedovSolution RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
```
jgphpc: There is a failure when building `sedovSolution` with clang++:
```
ld.lld: error: undefined symbol: std::filesystem::__cxx11::path::_M_split_cmpts()
>>> referenced by main.cpp
```

This fixes the issue:
```suggestion
target_link_libraries(sedovSolution stdc++fs)
install(TARGETS sedovSolution RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
```

jaescartin1:
Hi @sebkelle1 ,
     I set the iteration value to -1. Like first is check if time>0.0 (default value time=0.0),  if you don't specify time and you give the 'constant.txt' file instead but without iteration you will receive an error.
It make sense to check the iteration=0, if you want to compare the model at time=0. with the analytical solutión
... but I think there is an issue in the 'constants.txt' .  
iter=0 is the initial snapshot, without any movement but the constants file said that the time is t=dt, instead t=0.
Maybe you added t=t+dt before to write the line in the constants file.

jaescartin1:
$ cat constants.txt 
0 1.1e-06 1.1e-06 1 1.00739e-08 1 0 12702448 
1 2.31e-06 1.21e-06 1 3.10683e-08 1 0 12697729 
...

jaescartin1:
All the changes proposed in your comments are in-place.
Additionally I added the L1 error control for the common variables: rho, u, p, vel, cs
Please, review again the changes

jaescartin1:
Here the execution commands for the new version. 
All the output files will be generated in ./bin/
The pictures will be in .png format.
The call also check the rho L1 error

$ git clone --branch "analytical_sedov" https://github.com/jaescartin1/SPH-EXA.git
$ cd SPH-EXA && make -j mpi+omp && ./bin/mpi+omp.app -n 50 -s 200 -w 200 --outDir ./bin/
$ python src/analytical_solutions/compare_solutions.py sedov --binary_file bin/sedov_solution --constants_file ./bin/constants.txt --iteration 200 --nparts 125000 --snapshot_file ./bin/dump_sedov200.txt --out_dir bin/ --error_rho --delta_rho 1.e-0

jgphpc:
> I cannot see the python file in the build/ folder. Am I doing something wrong?

no, it will copy it only when running `make install`.
That said, I need to fix an issue in cmake (I'll push).

Btw, `-DBUILD_ANALYTICAL=ON` is on by default.

jgphpc:
cscs-ci run

sekelle:
cscs-ci run

sekelle:
This is ready for merging from my side. As soon as the CI is available again we can merge.

src/analytical_solutions/sedov_solution/sedov_solution.hpp
```diff
@@ -41,6 +41,9 @@
  *
  */
 
+#ifndef _SEDOV_SOLUTION_HPP_
+#define _SEDOV_SOLUTION_HPP_
+
```
sekelle: We don't need these include guards. We have "pragma once" just below

src/analytical_solutions/sedov_solution/particle.hpp
```diff
@@ -0,0 +1,62 @@
...
+/*! @file
+ *
+ * @brief Struct generated to sort the particles by radius
+ *
+ * @author Jose A. Escartin <ja.escartin@gmail.com>
+ *
+ */
+
+#ifndef _SEDOV_SOLUTION_PARTICLE_HPP_
+#define _SEDOV_SOLUTION_PARTICLE_HPP_
```
sekelle: not needed

src/analytical_solutions/sedov_solution/io.hpp
```diff
@@ -29,12 +29,17 @@
  *
  */
 
+#ifndef _SEDOV_SOLUTION_IO_HPP_
+#define _SEDOV_SOLUTION_IO_HPP_
```
sekelle: not needed

src/analytical_solutions/sedov_solution/main.cpp
```diff
@@ -0,0 +1,221 @@
...
+/*! @file
+ *
+ * @brief This program generate the analytical sedov solution based in the time and the initial conditions
+ *
+ * @author Jose A. Escartin <ja.escartin@gmail.com>
+ *
+ */
+
+#include <string>
+#include <iostream>
```
jgphpc, line 33:
```suggestion
#include <sstream>
#include <iostream>
```

src/analytical_solutions/sedov_solution/io.cpp
```diff
@@ -0,0 +1,200 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include "io.hpp"
+
+#include <iostream>
```
jgphpc, line 26:
```suggestion
#include <sstream>
#include <iostream>
```

src/analytical_solutions/sedov_solution/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+set(SOURCES io.cpp sedov_solution.cpp main.cpp)
+add_executable(sedov_solution ${SOURCES})
+target_include_directories(sedov_solution PRIVATE ${PROJECT_SOURCE_DIR}/include)
+target_include_directories(sedov_solution PRIVATE ${PROJECT_SOURCE_DIR}/src)
+target_link_libraries(sedov_solution stdc++fs)
```
jgphpc:
```suggestion
if(CMAKE_OSX_SYSROOT)
    target_link_libraries(sedov_solution stdc++)
else()
     target_link_libraries(sedov_solution stdc++fs)
endif()
```

jaescartin1:
Ruben is checking the Error L1 in u and cs. Let's wait for the merge till Ruben give the ok

jgphpc:
⚠️: I pushed compare_solutions.py (reformatting only, no change to the code)

jaescartin1:
Perfect, I put my last changes in your reformatted code

sekelle (approved):

jgphpc (approved):

PR #151, opened 2022-02-04
Analytical noh

jaescartin1:
- Noh test implemented
- generic code about test in include
- added generic functions in the compare script


sekelle:
Note about formatting: we have a C++ style defined in `.clang-format` in the top-level project dir. This PR does not comply with this formatting and in fact reverses a lot of the formatting fixes I did in the previous PR. You can use the `clang-format` tool to automatically format your code.

jaescartin1:
ok, yesterday night, I took the last merge PR #149  as base develop branch to apply my changes in order to avoid conflicts. 
Maybe now something are in conflict/rolled back from your merged changes in develop branch this morning (PR#150)

As you suggests, I will use the tool 'clang-format' to format my codes in your required stantard.
But, In my personal opinion, the strict format code sometimes enter in conflict with the readability, and in consequence with the maintainability, of the code. 

I am traveling today and I will check it on Sunday afternoon-night to be compatible with your last changes

BTW clang is only to do the code compatible with Apple users? or give us another advantages?

sebkelle1:
> But, In my personal opinion, the strict format code sometimes enter in conflict with the readability, and in consequence with the maintainability, of the code.

Yes, it might not always be appealing to adhere to a fixed set of formatting rules, but in general the result is pretty good,
it's automatic, and for larger collaborative projects probably the only feasible option. Indeed, `clang-format` is the de-facto
standard for this job that many projects use, like `git` for version control. In my CLion IDE, I can format highlighted code
with a keyboard short-cut, in addition to applying it to the whole file as the `clang-format` tool does. We're not everywhere
100% compliant yet, but we're getting there. Remark: You can still influence the formatting a bit by choosing where to put
line breaks.

> BTW clang is only to do the code compatible with Apple users? or give us another advantages?
No, the Clang compiler and it's LLVM backend is now one of the most popular cross-platform open-source compilers.
If you log into a Cray machine for instance, you'll see that Cray compilers are now based on Clang. Same for the AMD HIP compilers.

sekelle:
> But, In my personal opinion, the strict format code sometimes enter in conflict with the readability, and in consequence with the maintainability, of the code.

Yes, it might not always be appealing to adhere to a fixed set of formatting rules, but in general the result is pretty good,
it's automatic, and for larger collaborative projects probably the only feasible option. Indeed, `clang-format` is the de-facto
standard for this job that many projects use, like `git` for version control. In my CLion IDE, I can format highlighted code
with a keyboard short-cut, in addition to applying it to the whole file as the `clang-format` tool does. We're not everywhere
100% compliant yet, but we're getting there. Remark: You can still influence the formatting a bit by choosing where to put
line breaks.

> BTW clang is only to do the code compatible with Apple users? or give us another advantages?

No, the Clang compiler and it's LLVM backend is now one of the most popular cross-platform open-source compilers.
If you log into a Cray machine for instance, you'll see that Cray compilers are now based on Clang. Same for the AMD HIP compilers.

sekelle:
Need to merge develop to resolve conflicts.

sekelle (dismissed):
Very nice work on the Noh test, all of the code in `analytical_solutions` is good to go in.

There are some design inconsistencies in the general part of the code:
- the directory name `test` or `tests` is reserved for unit testing, you should not put actual source code used in the SPH apps into a test directory
- the files `io.hpp` and `particle_io.hpp` belong in `src/analytical_solutions` because they are only used there. The code in those files is also not generally applicable and is not a candidate to be used anywhere else apart from the analytical solutions.
- `test_file_writer.hpp` also belongs back somewhere in `src`. Any i/o code that explicitly names the particle fields of `ParticlesData` is test-case specific and does not belong in `include`. You can see this in the existing ASCII output in `include/file_utils.hpp`: Note how the binary input reading and the ASCII output is generic. It works for any number of fields that the caller specified. Your ascii parser violates this principle of separating the data layout from i/o parsing.

src/analytical_solutions/sedov_solution/main.cpp
```diff
@@ -92,36 +85,47 @@ int main(int argc, char** argv)
...
+            vSim[i] = {i,
+                       r,vel,
+                       x[i],y[i],z[i],
+                       vx[i],vy[i],vz[i],
+                       h[i],
+                       rho[i],u[i],p[i],
+                       cs[i],
+                       Px[i],Py[i],Pz[i]};
+        }
+        sort(vSim.begin(),vSim.end(),ParticleIO::cmp());
```
sekelle: Side remark: sorting particle data according to some specified criterion is a standard problem in SPH and has a standard solution. We're doing this on every SPH step to sort the particles according to ascending space filling curve key. The approach taken here has the downside that you first have to define a custom data layout `struct ParticleIO` and copy the original data there before it can be sorted. You have to pay the memory space for a full duplication of all the particles data and also, if the number of particle fields change, you have to adapt your data layout.

A standard way of solving this problem involves using a function called `sort_by_key`, it sorts an array of "values" according to another array of provided "keys". In your example, the keys are the radii, and the values are an `std::iota` sequence (0, 1, 2, 3, 4, 5, ...). The effect is that the iota sequence will be changed into the ordering that sorted your keys. You can then apply this ordering to as many particle buffers as you like.

An example (`test/integration_mpi/domain_nranks.cpp`):
```c++
    computeSfcKeys(x.data(), y.data(), z.data(), sfcKindPointer(codesGlobal.data()), numParticles, box);
    std::vector<LocalIndex> ordering(numParticles);
    std::iota(begin(ordering), end(ordering), LocalIndex(0));
    sort_by_key(begin(codesGlobal), end(codesGlobal), begin(ordering));
    reorderInPlace(ordering, x.data());
    reorderInPlace(ordering, y.data());
    reorderInPlace(ordering, z.data());
    reorderInPlace(ordering, h.data());
```

`sort_by_key` is for instance supported by the `Thrust` library, which means that we can immediately also apply this approach to buffers on the GPU. In addition, we don't need a custom sort key and the sorting happens on elementary types, which means that the implementation of `sort_by_key` can use O(N) radix-sort instead of O(NlogN) merge sort.

src/analytical_solutions/sedov_solution/main.cpp
```diff
@@ -92,36 +85,47 @@ int main(int argc, char** argv)
...
+            h,
+            rho,u,p,
+            cs,
+            Px,Py,Pz);
+
+        // Calculate radius, velocity and sort particle data by radius
+        for(size_t i = 0; i < nParts; i++)
+        {
+            double r   = sqrt( pow(x[i], 2) + pow(y[i], 2) + pow(z[i], 2) );
+            double vel = sqrt( pow(vx[i],2) + pow(vy[i],2) + pow(vz[i],2) );
```
sekelle: `pow(x, 2)` is at least 100 times more expensive than `x*x`...

src/analytical_solutions/noh_solution/main.cpp
```diff
@@ -0,0 +1,219 @@
...
+    vector<double> vz (nParts);
+    vector<double> h  (nParts);
+    vector<double> rho(nParts);
+    vector<double> u  (nParts);
+    vector<double> p  (nParts);
+    vector<double> cs (nParts);
+    vector<double> Px (nParts);
+    vector<double> Py (nParts);
+    vector<double> Pz (nParts);
+    vector<ParticleIO> vSim(nParts);
```
sekelle: Why duplicate all these fields when you could just use `ParticlesData`? The data you're reading was created by a code that operated on `ParticlesData`, so if anything changes there, you'll have reflect the changes here (e.g. there's no guarantee that `double` will always be used for all fields).

sekelle:
cscs-ci run

jgphpc:
You may want to look at the [report](https://github.com/unibas-dmi-hpc/SPH-EXA/runs/5091792233?check_suite_focus=true) from clang-format (ignore the files not part of this PR).

jaescartin1:
I am working in the fixes during today

src/analytical_solutions/sedov_solution/main.cpp
jaescartin1: fixed

jaescartin1:
Ah, ok about clang, good to know. I don't have access to Cray machines and in my previous experiences only we used clang for cross-platform compatibility but only with Mac. I adapted the part that I usually touch to clang-format.

jaescartin1:
- I removed the folder include/tests. 
- I moved the common files about solutions to /scrc/analytical_solutions/common/
- I moved the common files about tests to /src/common/

jaescartin1:
Thanks jg,
    I used your changes for convert to clang-format on the files that usually modified.

I also modified the command in '.github/workflows/clang-format.yml' to apply only to c++ files in the whole code.
- From:  "clang-format -i --style=file `find src include domain -type f`"
- To: "clang-format -i --style=file `find . \( -name "*.hpp" -o -name "*.cpp" -o -name "*.cu" -o -name "*.cuh" \)`"

jaescartin1:
Hi Sebastian,
as you suggested, the codes for sph-exa and the solution are independent binaries.

I tried to isolate as much as possible these solutions from the original code but unfortunately, by design requirements, there are bigger links between them:
- 'src/.../...DataGenerator' classes to take the initial values for the test. It include the 'particle_data.cpp' by side effect on this purpose.
- Read SPH-EXA output files products, like input files for generate the solutions.

As you explained, it is not a good coding practice add more that necessary (and put them in the wrong places, you are right). 
By design, include this complexity (like the namespace sphexa), was necessary to access the initial values.

Profiting the existing link for the sphexa namespace, I also added some generic classes (without any other includes) in order to not duplicate code:
- 'include/arg_parser.hpp'

Create a simple 'ParticleIO' specifically for analytical_solutions were a system to simplify the readability and maintainability of the solutions. This action try to reduce the impact of the changes inside the SPH-EXA code only to read the right columns in the outputs.

Of course, I can use the class "sphexa::ParticlesData" and the system to read/write particles that you implemented in SPH-EXA. 
But it is clear that we need to pay a cost on this change increasing the link between both binaries.

jaescartin1:
Analytical_solutions with templates and clang-formatted

jaescartin1:
About the 'src/common/test_file_writer.hpp', please let me know if you wish this generic implementation or other. I am not sure if I got the right detail on your comment

I checked the SPH-EXA code before Chrismas (mini-app) and I only did a generalization of the subclass class SedovFileWriter ('src/sedov/SedovFileWriter.hpp'), where you implemented the marked '... const = 0'  heritage functions in the superclass IFileWriter ('include/ifile_writer.hpp')

jaescartin1:
Working in the request of use ParticleData and SPH-EXA sort in analytical_solution's.

src/analytical_solutions/noh_solution/main.cpp
jaescartin1: ParticleData integrated, only last sort by radius

src/analytical_solutions/sedov_solution/main.cpp
jaescartin1: The ParticleData is now integrated. I am checking how to sort.

jaescartin1: Reordering in place

jaescartin1:
All the changes are in place, please check it again

jgphpc:
cscs-ci run

include/sph/density.hpp
```diff
@@ -52,6 +52,7 @@ void computeDensityImpl(const Task& t, Dataset& d, const cstone::Box<T>& box)
                        : ro [:n])
 #pragma omp teams distribute parallel for
 // clang-format on
+#else
```
sekelle: good catch

jaescartin1: :)

jaescartin1:
I merged with develop after PR#161 Gravity, and I working in the fixes

jaescartin1:
I am done. Please, execute cscs-ci run

jaescartin1 (commented):
the changes are already there

jgphpc:
cscs-ci run

include/sph/equationOfState.hpp
```diff
@@ -27,7 +27,7 @@ void computeEquationOfState(size_t startIndex, size_t endIndex, Dataset& d)
     T* cv   = d.cv.data();
 
 #pragma omp parallel for schedule(static)
-    for (size_t i = 0; i < endIndex; ++i)
+    for (size_t i = startIndex; i < endIndex; ++i)
```
sekelle: Good catch.

domain/include/cstone/focus/octree_focus_mpi.hpp
```diff
@@ -138,7 +138,7 @@ class FocusedOctree
      */
     void updateCounts(gsl::span<const KeyType> particleKeys,
                       gsl::span<const int> peerRanks,
-                      const SpaceCurveAssignment& assignment,
+                      [[maybe_unused]] const SpaceCurveAssignment& assignment,
```
sekelle: I'll remove this in a separate PR.

domain/include/cstone/domain/assignment.hpp
```diff
@@ -98,7 +98,7 @@ class GlobalAssignment
 
         gsl::span<const KeyType> oldLeaves = tree_.treeLeaves();
         std::vector<KeyType> oldBoundaries(assignment_.numRanks() + 1);
-        for (int rank = 0; rank < oldBoundaries.size() - 1; ++rank)
+        for (KeyType rank = 0; rank < oldBoundaries.size() - 1; ++rank)
```
sekelle: `KeyType` is used exclusively for space filling curve keys, like Hilbert or Morton keys.

sekelle (changes requested):
With the flexible field-names based file I/O and HDF5 output and parsing now available (#173, #174 ) this can be cleaned up and reworked.
In general if you find and fix issues in unrelated parts of the code, you should do that in a separate PR.

include/sph/equationOfState.hpp
sekelle: (Note, your suggestion is incorrect, the loop has to go from `0` to `numParticles`, not just `endIndex`)
I'll also address this in a separate PR.

sekelle:
This is up to date and ready now. The Noh solution is calculated directly in python. For L1 error sums, no interpolation is used.

sekelle:
noh_density_0 282685
noh_energy_0 282685
noh_pressure_0 282685
noh_velocity_0 282685

sekelle:
cscs-ci run

sekelle (approved):

PR #166, opened 2022-02-17
Fix include fail for gcc 11.2

biddisco:


jgphpc:
done with https://github.com/unibas-dmi-hpc/SPH-EXA/pull/167

PR #170, opened 2022-02-18
Volume elements

rmcabezon:
It adds volume elements, artificial viscosity switches, and necessary calculations for those.


sekelle:
This will need manual intervention to integrate. I'll take care of this as soon as I'm done with improvements to ParticleData and file I/O.

sekelle:
Addressed by PR #183

PR #171, opened 2022-02-18
Replace CMAKE_INSTALL_SBINDIR with CMAKE_INSTALL_BINDIR

biddisco:
I got an error with install and replaced `CMAKE_INSTALL_SBINDIR` - which looks wrong to me. Is this a typo that crept into the code somehow?


biddisco:
Closing this because I have now found the docs that refer to sbindir and apparently its ok.

PR #174, opened 2022-02-22
Simplification of Sedov analytical solution

sekelle:
- We do not need to sort SPH particles by radius for plotting. The particles are drawn one-by-one, the order is irrelevant.
- The analytical solution program has no business reading SPH simulation particle data. Its only purpose is to compute the solution
- We do not need to compute particle radii in C++ and store them to file. We can do that directly in the python script.
- This change depends on PR #173, enabling HDF5 output in SPH-EXA which can easily be parsed in python with h5py.
- The resulting HDF5 file contains all output steps as well as simulation step time as attributes.
- Reduced maintenance burden of file I/O: We can add and subtract particle fields for file I/O in any order we want at runtime without any code changes or even recompilation. The only requirement is to have common field names, e.g. "x", "vx", "p".
This means we can dynamically restrict output to the minimum required for a certain plot, e.g. `./sedov -s 100 -w 5 -f x,y,z,ro` for a density profile.


jaescartin1:
- Point 1 of your comment)

Although, we don't need sort the particles, the order is relevant for the solution.
For the simulation it doesn't matter, but you are touching the solution ant it have another different purposes.
The solution comparison is used for user test and experiment, including the plotting and the check of differences.
The plots of the solution should be done in lines and the simulation in points.
... you know what happens if you plot with lines a graphics with and not sorted r in one axis. 

The solution is calculated in a few points (1000), in case you don't ask for check L1 differences.

If you calculate L1 errors: (sum_i=1_n abs(sim[i]-sol[i]))/n ; the solution is calculate in every single radius point.
If you ask for L1 errors and plot ... and you don't order the particles ... your plot will be a mess.


- Point 2 of your comment)

I am fully agree, in the analytical solutions part, we don't need the complexity of ParticleData and the sort system used in SPH-EXA.
We only needed sort the particles, and I was happy with my little 30-lines code with my little struct with doubles and the c++ standard mergesort.
... but you requested that changes: don't use another new structure, use template datatype, use particleData and use the same sort that in SPH-EXA. 
It took me quite time to adapt all of this requests.

In my opinion this design introduce a link to SPH-EXA that, I guess, you initially wanted to avoid when you requested to me an independent application.

As I explained in the point 1, the program not only calculate the solution, also prepare the comparison between the simulation ant the solution.
This is because, this solution could be used for automatic check, like jg did, and for regression tests in the future.
We should use this generic test (and other) with analytical solution to test the impact of the changes when we add new physics or versions of the physics.

- Point 3 of your comment)

This is not right, we need to calculate the radius for the simulation (only in the solution code, not in the SPH-EXA code) for compare the simulation with the solution.
The solution code, generate 2 files: one for the SPH-EXA simulation and another for the analytical solution, with similar columns to do the live of the user easy for compare.
As you can see, the solution generate 1D positions (only radii). 
To have the same fields we need to calculate for example the radius, and other variables or observable's that can be useful to compare simulation and solution.

- Point 4 and next's)

I asked previously to have the flexibility to change the outputs in the analytical solution part (I am not talking about I/O SPH-EXA). I think, you just removed that part.
This allow us to prepare the output columns for the comparison, the only dependency is how read the SPH-EXA simulation (1 single function). 
About hdf5, I consider this a a good idea, because the previous binary format don't give us too much compression and it is ad-hoc. Basically hdf5 is generic.

jaescartin1:
In resume, after my comments where I exposed my opinion, up to you for apply the changes that you consider ... it is completely your project not mine, but I am a quite busy person (as you and the others) and I am doing this for fun.
... sincerely, I am a little bit tired to spend my free time to change the code again and again because I don't have a clear specification of your wishes.
As you could see, I can adapt my code to all the fancy things, but I hate to work 2,3,... times in the same implementation (how I did in the analytical sedov and how I am doing in the analytical noh PR).

Please, clarify your requests with Ruben and the others and let me know clearly how implement the future changes. 
I won't like to change the same design again in a few weeks.

jaescartin1 (approved):

sekelle:
It is true that my earlier review comments about sorting did lead to additional work on your part that now cannot be used anymore. I did not realize from the start that sorting particles with ascending radius is not needed, for that I am sorry.

We really do not need to load particle data in the analytic solution code and sort the radii. For calculating L1 errors, we don't need to calculate the analytical solution _at the radius of every single particle_. Think of how many duplicate evaluations
of the solution that would incur! For any given particle at radius r, you can get the value of the solution be interpolating between
the 1000 steps where you evaluated it it.

Now we have a solution code that produces just the solution and does not also read SPH simulation data. We can
still do whatever comparison of solution to simulation you can think of.

Finally, just because some parts of the code got replaced, it doesn't mean that the earlier versions were not useful.
In fact, we would not have the current version if you hadn't implemented a first version. So your contribution is definitely
still useful and alive.

jaescartin1:
Hi Sebastian,
        ok, no problem. I know that the design is an iterative process, I only ask for more communication to avoid unnecessary work. Please merge, we will live with these changes and redesign the working. I think that I need to redo quite parts of the analytical noh, because this was part was implemented and improved on that version that move up and down since 1 month.

But again, we will have another different problems if I am not allow to reorder particles. Checking the radius positions 1 by 1 for L1 errors was mandatory for Ruben. A month ago I proposed to him use the 1000 steps with interpolation ... and he said definitely no, better calculate the solution in the real radius positions.  The reason behind is simple, we have an exact algorithm that calculate the exact solution in one point, why are we not  using if we have it?

Additional we need to find the two more near values to the particle radius, i-1 and i, instead of calculate the solution in that exact position.

Avoiding these two steps, it will be less computational expensive (but negligible in front of the simulation):
- calculate the solution in only a few points, as just the code did in January. 
- And ordering the simulation particles
... but you need to think about the additional value of these tasks in order to simplify the user evaluation of the results.

PR #181, opened 2022-03-01
fix for nvc++

jgphpc:
Fix to avoid 
```
main.cpp", line 134: error: namespace "std" has no member "setw"
std::setw(16),
```


jgphpc:
>fatal: missing blob object

Andreas confirmed this is an issue with the CI (not the code)

sekelle (approved):

PR #188, opened 2022-03-10
Unified application front-end

sekelle:
- unify all test cases in a single SPH application
- support for starting simulations from initial conditions given by files and full checkpointing capability


jgphpc (approved):
can be merged, I'll add hdf5 to the ci in a separate PR

src/sphexa/sphexa.cpp
```diff
@@ -114,29 +110,24 @@ int main(int argc, char** argv)
...
         {
-            fileWriter->dump(d, domain.startIndex(), domain.endIndex(), outDirectory + "dump_sedov");
+            fileWriter->dump(d, domain.startIndex(), domain.endIndex(), box, outFile);
         }
 
         if (d.iteration % 5 == 0) { viz::execute(d, domain.startIndex(), domain.endIndex()); }
     }
 
-    totalTimer.step("Total execution time of " + std::to_string(maxStep) + " iterations of Sedov");
+    totalTimer.step("Total execution time of " + std::to_string(maxStep - startIteration + 1) + " iterations of Sedov");
```
jgphpc: you may want to update the message when running `noh` instead of sedov ?

sekelle: done

PR #193, opened 2022-03-16
Radiative cooling

nknk567:


sph/include/sph/cooling.hpp
```diff
@@ -0,0 +1,141 @@
+//
```
jgphpc: need license here

extern/grackle_deps/src/clib/auto_show_version.c
```diff
@@ -0,0 +1,8 @@
+#include <stdio.h>
+void auto_show_version(FILE *fp) {
+   fprintf (fp,"\n");
+   fprintf (fp,"The Grackle Version xxx\n");
+   fprintf (fp,"Git Branch   master\n");
+   fprintf (fp,"Git Revision xxx\n");
+   fprintf (fp,"\n");
+}
```
sekelle: Perhaps the 3 files in this folder could be combined in a single file `show_config.c`?

sph/include/sph/cooling.hpp
```diff
@@ -0,0 +1,141 @@
...
+    d.e_fraction[i]     = e_density / gr_rho;
+    d.HM_fraction[i]    = HM_density / gr_rho;
+    d.H2I_fraction[i]   = H2I_density / gr_rho;
+    d.H2II_fraction[i]  = H2II_density / gr_rho;
+    d.DI_fraction[i]    = DI_density / gr_rho;
+    d.DII_fraction[i]   = DII_density / gr_rho;
+    d.HDI_fraction[i]   = HDI_density / gr_rho;
+    d.metal_fraction[i] = metal_density / gr_rho;
+    d.u[i]              = gr_u;
+}
```
sekelle: Does ChanGa also convert between fractions and densities every particle and iteration?

sph/include/sph/cooling.hpp
```diff
@@ -0,0 +1,141 @@
...
+    gr_float hr                            = 0.;
+    grackle_fields.volumetric_heating_rate = &vhr;
+    grackle_fields.specific_heating_rate   = &shr;
+    grackle_fields.RT_HI_ionization_rate   = &HI_ir;
+    grackle_fields.RT_HeI_ionization_rate  = &HeI_ir;
+    grackle_fields.RT_HeII_ionization_rate = &HeII_ir;
+    grackle_fields.RT_H2_dissociation_rate = &H2_dr;
+    grackle_fields.RT_heating_rate         = &hr;
+    gr_float isrf                          = 1.7;
+    grackle_fields.isrf_habing             = &isrf;
```
sekelle: Do these configuration fields stay the same for all particles and even between iterations?

sph/include/sph/cooling.hpp
```diff
@@ -0,0 +1,141 @@
...
+}
+#include <cmath>
+static constexpr double m_sun = 1.989e33;
+static constexpr double pc    = 3.086e18;
+static constexpr double G_cgs = 6.674e-8;
+
+static code_units grackle_units;
+
+template<typename T, typename Dataset>
+void cool_particle(Dataset& d, size_t i)
```
sekelle: This function receives way too much information compared to what it is actually using.
The arguments contain information of _all_ particle fields for _all_ particles, even though
this is only using the composite fractions and rho and u for a _single_ particle.
I would propose you pass here only the pointers to rho[i], u[i],HI[i], etc.

sekelle (changes requested):

sekelle:
Since radiative cooling is not really related to hydrodynamics, I would propose we introduce a new top-level directory `cooling`.
We can leave Grackle at `extern/grackle` as the git submodule and in `cooling` we have `CMakeLists.txt`, `src` with the wrapper function by Noah and show_config.c and `test` for the unit test.
The dependency on `ParticlesData` will go away with the suggestion above. This will make the cooling module self-contained and can be tested without sphexa includes. SPH-EXA should depend on `cooling`, not `cooling` on SPH-EXA.

sph/include/sph/cooling.hpp
nknk567: Yes.

sph/include/sph/cooling.hpp
nknk567: Yes, I will change that.

sph/include/sph/cooling.hpp
nknk567: No, but they are optional and not implemented in changa. If we want them, they probably also have to be added to the dataset.

sph/include/sph/cooling.hpp
nknk567: I copied the same license as is in the other files.

rmcabezon:
> Since radiative cooling is not really related to hydrodynamics, I would propose we introduce a new top-level directory `cooling`. We can leave Grackle at `extern/grackle` as the git submodule and in `cooling` we have `CMakeLists.txt`, `src` with the wrapper function by Noah and show_config.c and `test` for the unit test. The dependency on `ParticlesData` will go away with the suggestion above. This will make the cooling module self-contained and can be tested without sphexa includes. SPH-EXA should depend on `cooling`, not `cooling` on SPH-EXA.

We might find the same issue with nuclear reactions. Why not have a "physics" top-level directory and inside sub-directories with different physical processes?

sekelle (approved):

PR #195, opened 2022-03-17
Evrard glass sphere

sekelle:


jgphpc (approved):

jaescartin1 (approved):

main/src/init/noh_init.hpp
```diff
@@ -42,15 +42,12 @@ namespace sphexa
...
 {
     using T = typename Dataset::RealType;
 
-    int    ng0         = 100;
-    double r1          = constants.at("r1");
-    double totalVolume = std::pow(2 * r1, 3);
-    double hInit       = std::cbrt(3.0 / (4 * M_PI) * ng0 * totalVolume / d.numParticlesGlobal) * 0.5;
-
+    int    ng0           = 100;
+    double hInit         = std::cbrt(3.0 / (4 * M_PI) * ng0 * totalVolume / d.numParticlesGlobal) * 0.5;
```
rmcabezon, line 50: This formula is assuming constant density. It is more general to use local values: i.e. hInit = std::cbrt(3 / 4 / PI * ng0 * m[i] / rho[i]) * 0 .5

rmcabezon: Which for Noh is true, but better to keep it general.

rmcabezon (commented):
I suggested a change in the h calculation

main/src/init/noh_init.hpp
sekelle: For one, this is not general code but Noh-specific initialization.
Two, we don't have rho[i] at initialization time.

PR #203, opened 2022-03-30
Added a observables interface and growth rate calculations for the Kelvin-Helmholtz test case

lks1248:


main/src/sphexa/sphexa.cpp
```diff
@@ -122,7 +125,7 @@ int main(int argc, char** argv)
 
         if (rank == 0)
         {
-            fileutils::writeColumns(constantsFile, ' ', d.iteration, d.ttot, d.minDt, d.etot, d.ecin, d.eint, d.egrav);
+            observables -> computeAndWrite(d, domain.startIndex(), domain.endIndex(), box);
         }
```
sekelle: `if (rank ==0)` must be removed here as the observables call has to be executed by all ranks.

main/src/io/time_energy_growth.hpp
```diff
@@ -0,0 +1,129 @@
...
+    TimeEnergyGrowth(std::ofstream& constPath) : constantsFile(constPath){}
+
+    using T = typename Dataset::RealType; 
+    void computeAndWrite(Dataset& d, size_t firstIndex, size_t lastIndex,
+                    cstone::Box<T>& box)
+    {
+        T khgr = computeKHGrowthRate<T>(firstIndex, lastIndex, d, box);
+        printf("KH Growth Rate this iteration: %f\n", khgr);
+        fileutils::writeColumns(constantsFile, ' ', d.iteration, d.ttot, d.minDt, d.etot, d.ecin, d.eint, d.egrav, khgr);
+    }
```
sekelle: only rank 0 should write into the file

main/src/io/time_energies.hpp
```diff
@@ -0,0 +1,64 @@
...
+public:
+
+    TimeAndEnergy(std::ofstream& constPath) : constantsFile(constPath){}
+
+    using T = typename Dataset::RealType; 
+    void computeAndWrite(Dataset& d, size_t firstIndex, size_t lastIndex,
+                    cstone::Box<T>& box)
+    {
+        fileutils::writeColumns(constantsFile, ' ', d.iteration, d.ttot, d.minDt, d.etot, d.ecin, d.eint, d.egrav);
+    }
```
sekelle: only rank 0 should write the file

main/src/io/time_energy_growth.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+/*! @file
+ * @brief output and calculate energies and growth rate for Kelvin-Helmholtz tests
+ *
+ * @author Lukas Schmidt
+ */
+
+#include "file_utils.hpp"
+#include "iobservables.hpp"
+#include "sph/math.hpp" //??
```
sekelle: can be removed from the include list?

main/src/io/time_energy_growth.hpp
```diff
@@ -0,0 +1,129 @@
...
+    MPI_Reduce(sum, globalSum, 3, MpiType<T>{}, MPI_SUM, rootRank, MPI_COMM_WORLD);
+#endif
+    return 2.e0 * std::sqrt((globalSum[0]/globalSum[2])*(globalSum[0]/globalSum[2]) + (globalSum[1]/globalSum[2])*(globalSum[1]/globalSum[2]));
+    ;
+}
+
+
+
+
+template<class Dataset>
```
sekelle: needs a @brief description

main/src/io/time_energy_growth.hpp
```diff
@@ -0,0 +1,129 @@
...
+#include "file_utils.hpp"
+#include "iobservables.hpp"
+#include "sph/math.hpp" //??
+#include <fstream>
+#include "ifile_writer.hpp"
+
+
+namespace sphexa {
+
+template<typename T, class Dataset>
```
sekelle: needs a docstring (what is this function doing?)

main/src/io/factory.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include "time_energy_growth.hpp"
+#include "time_energies.hpp"
+
+namespace sphexa
+{
+
+template<class Dataset>
+std::unique_ptr<IObservables<Dataset>> observablesFactory(std::string testCase, std::ofstream& constantsFile)
+{
+    if (testCase == "IC_kelvin-helmholtz.h5")
```
sekelle: This is one ugly hack :)
One possible way to make this nicer is to set a boolean file attribute in the input hdf5 file that contains the ICs. Something like "KelvinHelmoltzGrowthRate=true/false".
You can use std::filesystem to check whether `testCase` is an existing file. If yes, you next get the list of file attributes. If that list contains KelvinHelmholtzGrowthRate and the  value is true, you select the Observable here.

main/src/io/time_energy_growth.hpp
```diff
@@ -0,0 +1,129 @@
...
+T computeKHGrowthRate(size_t startIndex, size_t endIndex, Dataset& d, const cstone::Box<T>& box)
+{
+    T sum[3], globalSum[3];
+    localGrowthRate(startIndex, endIndex, d, sum + 0, sum + 1, sum + 2, box);
+
+    int rootRank = 0;
+#ifdef USE_MPI
+    MPI_Reduce(sum, globalSum, 3, MpiType<T>{}, MPI_SUM, rootRank, MPI_COMM_WORLD);
+#endif
```
sekelle: you can omit the #ifdefs here, we anyway have MPI mandatory already.

sekelle (changes requested):
This is in very good shape already. The interfaces and organization of the additional observable code are spot-on.

Some changes are needed to make this work on multiple ranks.
You can for instance pass the MPI communicator to the constructor together with the constants file name. The communicator can be used to figure out the executing rank and can be used for the collective reduction. Thanks to knowing the rank ID you can make sure that only rank 0 writes the results to file.

From an organizational point of view, this code is not really doing i/o, the files here can go into a new directory main/src/observables

main/src/io/time_energy_growth.hpp
rmcabezon: Somewhere I would add that this calculation and the growth rate are taken from McNally et al. ApJSS, 201 (2012)

rmcabezon (approved):

sekelle:
cscs-ci run

sekelle:
There are some compilation errors due to missing include statements:
https://gitlab.com/cscs-ci/ci-testing/webhook-ci/SPH-EXA-44692846847247/-/jobs/2304264659

Could you use `std::string` instead of `char*` for string comparison?

main/src/observables/factory.hpp
```diff
@@ -0,0 +1,87 @@
...
+    {
+        H5PartFile* h5_file = nullptr;
+        h5_file = H5PartOpenFile(testCase.c_str(), H5PART_READ);
+        size_t attrN = H5PartGetNumFileAttribs(h5_file);
+        char* attrName = new char; 
+        long int length = 40; //arbitrary high value to read enough
+        long int* attr_type;
+        long int* attr_nelem;
+
+
```
sekelle: This code actually has several issues:

- `char* attrName = new char; ` create pointer to just one, not 40, char on the heap
- The `H5PartGetFileAttribInfo` then writes past the bounds because `attrName` is length 1, not 40
- `attrName` is not deallocated, you've got a memory leak here.

main/src/observables/time_energy_growth.hpp
```diff
@@ -0,0 +1,144 @@
...
+            T aux = std::exp(-4.e0 * PI * std::abs(y[i] - 0.25e0));
+            si = vy[i] * voli * std::sin(4.e0 * PI * x[i]) * aux;
+            ci = vy[i] * voli * std::cos(4.e0 * PI * x[i]) * aux;
+            di = voli * aux;
+        } else {
+            T aux = std::exp(-4.e0 * PI * std::abs((ybox - y[i]) - 0.25e0));
+            si = vy[i] * voli * std::sin(4.e0 * PI * x[i]) * aux;
+            ci = vy[i] * voli * std::cos(4.e0 * PI * x[i]) * aux;
+            di = voli * aux;
+        }
```
sekelle, line 67: In this if/else clause, only the definition of `aux` differs, the other three statements are identical and can be taken outside, if not a bug.

main/src/observables/time_energy_growth.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+namespace sphexa {
+
+
+template<typename T, class Dataset>
+void localGrowthRate(size_t startIndex, size_t endIndex, Dataset& d, T* sumsi, T* sumci, T* sumdi, const cstone::Box<T>& box)
+{
+    const T* x  =   d.x.data();
+    const T* y  =   d.y.data();
+    const T* vx =   d.vx.data();
```
sekelle: `vx` is never accessed. Maybe a bug?

sekelle:
I fixed the HDF5 attribute part. The only thing left to address is to verify the comments about the growth rate calculation.

sekelle:
cscs-ci run

main/src/observables/time_energy_growth.hpp
lks1248: you are right, they are the same. fixed in the latest commit

main/src/observables/time_energy_growth.hpp
lks1248: resolved in the latest commit :)

sekelle:
cscs-ci run

sekelle (approved):

PR #206, opened 2022-04-01
Fix compilation with serial HDF5

jgphpc:


main/src/io/mpi_file_utils.hpp
```diff
@@ -62,6 +62,8 @@ void writeH5Part(Dataset& d, size_t firstIndex, size_t lastIndex, const cstone::
     // output name
     const char* h5_fname = path.c_str();
     H5PartFile* h5_file  = nullptr;
+    int         numRanks;
+    MPI_Comm_size(d.comm, &numRanks);
```
sekelle: This can be moved down into the `#else` section.

sekelle (approved):

PR #207, opened 2022-04-01
Include git info in the sphexa binary

jgphpc:


main/src/CMakeLists.txt
```diff
@@ -2,14 +2,13 @@
...
-    endif ()
-endfunction()
-
+# function(enableH5Part exename)
+#     if (SPH_EXA_WITH_H5PART)
+#         target_compile_definitions(${exename} PUBLIC SPH_EXA_HAVE_H5PART)
+#         target_include_directories(${exename} PUBLIC ${PROJECT_SOURCE_DIR}/extern/h5part)
+#         target_link_libraries(${exename} PRIVATE H5Part ${HDF5_LIBRARIES})
+#     endif ()
+# endfunction()
```
sekelle: Can be deleted?

main/src/sphexa/CMakeLists.txt
```diff
@@ -34,6 +58,7 @@ endif()
 if(CMAKE_CUDA_COMPILER)
     add_executable(${exename}-cuda $<TARGET_OBJECTS:gather_obj> $<TARGET_OBJECTS:cuda_find_neighbors_obj> $<TARGET_OBJECTS:cuda_sph> sphexa.cpp)
     target_include_directories(${exename}-cuda PRIVATE ${SPH_EXA_INCLUDE_DIRS})
+    target_include_directories(${exename}-cuda PUBLIC ${CMAKE_BINARY_DIR}/main/src/util)
```
sekelle: I would choose `main/src/` as the location here to be consistent with the source tree

main/src/util/utils.hpp
```diff
@@ -2,6 +2,7 @@
 
 #include <tuple>
 #include <omp.h>
+#include <version.h>
```
sekelle: use quotes here, "version.h". The <> syntax is used for system and external library includes.

PR #213, opened 2022-04-08
Add command line parameter to gravity_perf test

jgphpc:


ryoanji/test/cpu/gravity.cpp
```diff
@@ -39,17 +39,18 @@
 
 using namespace ryoanji;
 
-int main()
+// int main()
```
sekelle:
```suggestion
```

sekelle (changes requested):

sekelle (approved):

PR #215, opened 2022-04-12
Gpu fields

sekelle:
Some cleanup and simplification related to gravity on the GPU.


sekelle:
CI is down and JG on holidays, performed checks manually.

PR #221, opened 2022-04-15
Compute minimum timestep without storing maxvsignal

sekelle:


sekelle:
cscs-ci run

sekelle:
cscs-ci run

PR #222, opened 2022-04-19
Domain fixes for Evrard

sekelle:
Various fixes to address problems with large-scale runs of the Evrard test case:
- Made domain exchange tolerate calls to `domain::sync(Grav)` with zero particles on some ranks
- Initialize `FocusOctree` with the same MAC criterion that is used for gravity traversal instead of minDistance MAC, this prevents MAC failures for far-way nodes that would lead to a catastrophic amount of halo particles in the first iteration.
- When gravity is active, also initialize domain with `domain::syncGrav` instead of `domain::sync` to ensure initial convergence of `FocusOctree` with the correct MAC criterion
- Change peer rank detection to a MAC criterion to a stricter combination of minDistance MAC and VecMac.
- Global focus exchange: `FocusOctree` has to tolerate a global tree with a higher resolution than the focus tree during initial build up
- Fixed `Octree:locate()` which was directly returning the result of `std::lower_bound` instead of checking whether the queried node exists in the tree.

### DIFF
--- a/domain/include/cstone/halos/halos.hpp
+++ b/domain/include/cstone/halos/halos.hpp
@@ -47,7 +47,8 @@ template<class KeyType>
 class Halos
 {
 public:
-    Halos(int myRank) : myRank_(myRank)
+    Halos(int myRank)
+        : myRank_(myRank)
     {
     }
 
@@ -153,7 +154,7 @@ private:
 
         for (int range = 0; range < 2; ++range)
         {
-            #pragma omp parallel for
+#pragma omp parallel for
             for (TreeNodeIndex i = checkRanges[range][0]; i < checkRanges[range][1]; ++i)
             {
                 if (haloFlags_[i])
@@ -161,10 +162,7 @@ private:
                     bool peerFound = false;
                     for (auto peerRange : focusAssignment)
                     {
-                        if (peerRange.start() <= i && i < peerRange.end())
-                        {
-                            peerFound = true;
-                        }
+                        if (peerRange.start() <= i && i < peerRange.end()) { peerFound = true; }
                     }
                     if (!peerFound)
                     {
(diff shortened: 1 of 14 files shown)


domain/include/cstone/halos/halos.hpp
```diff
@@ -173,7 +171,7 @@ class Halos
                                   << " of similar magnitude than the rank domain volume. In that case, either"
                                   << " the number of ranks needs to be decreased or the number of particles"
                                   << " increased, leading to shorter smoothing lengths\n";
-                        MPI_Abort(MPI_COMM_WORLD, 1);
+                        MPI_Abort(MPI_COMM_WORLD, 35);
```
jaescartin1, line 174: Is this MPI_Abort to catch some lost rank? 
Is 35 a random value, or only to differentiate the exit value to something else?

domain/include/cstone/traversal/peers.hpp
```diff
@@ -119,35 +119,36 @@ template<template<class> class TreeType, class KeyType, class T>
 std::vector<int> findPeersMacStt(
     int myRank, const SpaceCurveAssignment& assignment, const TreeType<KeyType>& octree, const Box<T>& box, float theta)
 {
-    float invTheta = 1.0f / theta;
+    float invTheta      = 1.0f / theta;
```
jaescartin1, line 122: Should this variable be float, double or generic?

jaescartin1 (approved):

domain/include/cstone/halos/halos.hpp
sekelle: I observed this MPI_Abort statement being triggered without getting the error print message above. That's why I'm
giving a unique return value (35) here to faster determine where the Abort came from, even if the error is not printed.

domain/include/cstone/traversal/peers.hpp
sekelle: we currently have theta as `float` everywhere

domain/include/cstone/halos/halos.hpp
jaescartin1: ok, thanks

domain/include/cstone/traversal/peers.hpp
jaescartin1: I see

PR #223, opened 2022-04-19
Output wextra

jaescartin1:
Add a new option --wextra that allow the user to introduce a list of times where the simulation is write to the disk


sekelle:
cscs-ci run

sekelle:
cscs-ci run

sekelle (approved):

PR #224, opened 2022-04-20
Fix h5 part warnings

jaescartin1:
Fix the -Wformat-overflow= warnings into the comilation


sekelle:
I'd rather not change any code in extern/ unless I have to due to compiler errors.
We can eliminate all the warnings in CMake by simply qualifying the h5part library as "SYSTEM"

jaescartin1:
ah, ok. If you can do that perfect. I close the PR

PR #226, opened 2022-04-20
Added float options to the input parameters -s and -w

jaescartin1:
-s (float) : maximum model time to simulate
-w (float) : frequency time to output the simulation


main/src/io/arg_parser.hpp
```diff
@@ -89,9 +91,31 @@ bool isExtraOutputStep(size_t step, double t1, double t2, const std::vector<std:
...
+ *
+ * @param step          number of simulation step
+ * @param frequency     iteration frequency to output the simulation (>0)
+ * @return              true if the step fit with the usual iteration frequency output
+ */
+bool isPeriodicOutputStep(size_t step, int frequency)
 {
-    return writeFrequency == 0 || (writeFrequency > 0 && step % writeFrequency == 0);
+    assert(frequency > 0);
+    return (step % frequency == 0);
```
sekelle:
```suggestion
    return frequency != 0 && (step % frequency == 0);
```
If you allow an input of zero and return the correct answer for that, then you don't need a conditional check before calling this function to avoid calling with zero.

main/src/sphexa/sphexa.cpp
```diff
@@ -129,7 +134,7 @@ int main(int argc, char** argv)
     bool  haveGrav = (d.g != 0.0);
     float theta    = parser.exists("--theta") ? parser.get<float>("--theta") : (haveGrav ? 0.5f : 1.0f);
 
-    if (rank == 0 && (writeFrequency > 0 || !writeExtra.empty()))
+    if (rank == 0 && (writeFrequency > 0. || !writeExtra.empty()))
```
sekelle: Floating point comparisons can be tricky.
```suggestion
    if (rank == 0 && (writeFrequencyStr != "0" || !writeExtra.empty()))
```

main/src/io/arg_parser.hpp
```diff
@@ -89,9 +91,31 @@ bool isExtraOutputStep(size_t step, double t1, double t2, const std::vector<std:
...
+ * @param frequency     frequency time to output the simulation (>0.)
+ * @return              true if the step have a time that fit with the usual time frequency output in [t, t+dt]
+ */
+bool isPeriodicOutputTime(double t, double dt, double frequency)
+{
+    assert(frequency > 0.);
+    double currentTime = std::fmod(t, frequency);
+    double futureTime  = currentTime + dt;
+    return (currentTime < dt) || ((futureTime > frequency) && (frequency - currentTime < futureTime - frequency));
+}
```
sekelle, line 108: same applies here, if `frequency` is zero, you simply return false.

main/src/io/arg_parser.hpp
```diff
@@ -89,9 +91,31 @@ bool isExtraOutputStep(size_t step, double t1, double t2, const std::vector<std:
...
 }
 
-bool isPeriodicOutputStep(size_t step, int writeFrequency)
+/*! @brief Evaluate if the current step should be output (to file) per time frequency
+ *
+ * @param t             simulation time at beginning of current step
+ * @param dt            time step of the step
+ * @param frequency     frequency time to output the simulation (>0.)
+ * @return              true if the step have a time that fit with the usual time frequency output in [t, t+dt]
+ */
```
sekelle, line 100: This function needs a small unit test. You may follow the example for `isExtraOutputStep` in `main/src/test/io/arg_parser.cpp`

main/src/sphexa/sphexa.cpp
```diff
@@ -156,23 +161,36 @@ int main(int argc, char** argv)
...
+    if (iMaxStep)
+    {
+        totalTimer.step("Total execution time of " + std::to_string(std::stoi(maxStepStr) - startIteration + 1) +
+                        " iterations of " + initCond);
+    }
+    else
+    {
+        totalTimer.step("Total execution time until " + std::to_string(std::stod(maxStepStr)) + " time simulation of " +
+                        initCond);
+    }
```
sekelle: We can combine these two branches into a single statement by saying:
```c++
totalTimer.step("Total execution time of " + std::to_string(std::stoi(maxStepStr) - startIteration + 1) +
                  " iterations of " + initCond + " up to t=" + std::to_string(d.ttot));
```

main/src/sphexa/sphexa.cpp
```diff
@@ -156,23 +161,36 @@ int main(int argc, char** argv)
...
+                                                      : isPeriodicOutputTime(d.ttot, d.minDt, writeFrequency);
+
+        if (writeXFrequency || isExtraOutputStep(d.iteration, d.ttot - d.minDt, d.ttot, writeExtra))
         {
             fileWriter->dump(d, domain.startIndex(), domain.endIndex(), box, outFile);
         }
 
         if (d.iteration % 50 == 0) { viz::execute(d, domain.startIndex(), domain.endIndex()); }
+
+        if ((iMaxStep && d.iteration == std::stoi(maxStepStr)) || (!iMaxStep && d.ttot > std::stod(maxStepStr))) break;
```
sekelle: Here we could have a small function `stopSimulation(d.iteration, d.ttot, maxStepStr)` to take the decision.
This would also make it possible to move `iMaxStep` in there.

main/src/io/arg_parser.hpp
```diff
@@ -89,9 +91,31 @@ bool isExtraOutputStep(size_t step, double t1, double t2, const std::vector<std:
...
+bool isPeriodicOutputTime(double t, double dt, double frequency)
+{
+    assert(frequency > 0.);
+    double currentTime = std::fmod(t, frequency);
+    double futureTime  = currentTime + dt;
```
sekelle: There's a subtle complication here: When this function is called, the current time is `d.ttot` and at the beginning of
the timestep, it was `d.ttot - d.minDt`. These are the two times that we have access to.
`d.ttot + d.minDt` is not guaranteed to be the simulation time of the next time step, because the dt that is added to `d.ttot` will be calculated in the next time step.

main/src/sphexa/sphexa.cpp
```diff
@@ -156,23 +161,36 @@ int main(int argc, char** argv)
...
-        observables -> computeAndWrite(d, domain.startIndex(), domain.endIndex(), box);
+        observables->computeAndWrite(d, domain.startIndex(), domain.endIndex(), box);
 
-        if (isPeriodicOutputStep(d.iteration, writeFrequency) ||
-            isExtraOutputStep(d.iteration, d.ttot - d.minDt, d.ttot, writeExtra))
+        bool writeXFrequency = (writeFrequency <= 0.) ? false
+                               : (iWriteFrequency)    ? isPeriodicOutputStep(d.iteration, writeFrequency)
+                                                      : isPeriodicOutputTime(d.ttot, d.minDt, writeFrequency);
+
+        if (writeXFrequency || isExtraOutputStep(d.iteration, d.ttot - d.minDt, d.ttot, writeExtra))
```
sekelle: Together with the other comments, it should be possible to write here
```c++
if (isPeriodicOutputStep(d.iteration, writeFrequencyStr) || isPeriodicOutputTime(d.ttot, d.minDt, writeFrequencyStr) ||
  isExtraOutputStep(...))
```

sekelle (changes requested):

main/src/sphexa/sphexa.cpp
jaescartin1: If maxStepStr(int) means iterations will work.
Maybe I am wrong but, if maxStepStr(float) means time, we are going to substract (time - interations + 1)

sekelle: Yes true, but there's an easy fix: use `d.iteration` instead of `maxStepStr`.

jaescartin1: ok, if you don't need this difference for me it is fine

sekelle: By "difference" you mean whether number of steps or time limit caused the simulation to stop?

jaescartin1: The info about the exit condition in the stop simulation. If it doesn't matter show this info, it will be fine

main/src/sphexa/sphexa.cpp
jaescartin1: moved to one function

jaescartin1:
Can you do the other changes yourself? I think they are simple changes and it will be faster.

sekelle:
cscs-ci run

sekelle (approved):

PR #227, opened 2022-04-21
Fix HDF5 attribute check and clean KH growth rate

sekelle:


main/src/observables/time_energy_growth.hpp
```diff
@@ -30,78 +30,74 @@
...
+{
+    const Tc ybox = box.ly();
 
+    Tc sumsi = 0;
+    Tc sumci = 0;
+    Tc sumdi = 0;
 
-template<typename T, class Dataset>
-void localGrowthRate(size_t startIndex, size_t endIndex, Dataset& d, T* sumsi, T* sumci, T* sumdi, const cstone::Box<T>& box)
-{
```
sekelle, line 44: This function only needs arrays for `x, y, vy, rho, m, kx`, so we pass that directly instead of the whole Dataset,
achieving better code modularity. If we want to write a test for this, we now only have to supply the needed arrays,
rather than create an entire Dataset.

main/src/observables/time_energy_growth.hpp
```diff
@@ -30,78 +30,74 @@
...
-    const T* x  =   d.x.data();
-    const T* y  =   d.y.data();
-    const T* vy =   d.vy.data();
-    const T* rho =  d.rho.data();
-    const T* m =    d.m.data();
-    const T* kx =   d.kx.data();
-    const T ybox =  box.ly();
-
-    T sumsiThread = 0.0, sumciThread = 0.0, sumdiThread = 0.0;
```
sekelle, line 53: These variables are defined outside the OpenMP clause and are therefore not thread-local.
OpenMP will automatically generate thread-local variables for the reduction.

main/src/observables/time_energy_growth.hpp
```diff
@@ -30,78 +30,74 @@
...
-        sumciThread += ci;
-        sumdiThread += di;
+        sumsi += si;
+        sumci += ci;
+        sumdi += di;
     }
 
-    *sumsi = sumsiThread;
-    *sumci = sumciThread;
-    *sumdi = sumdiThread;
```
sekelle, line 79: Returning simple variables with pointers is old C-style, not modern C++.

lks1248 (commented):
Looks good to me. I think it would be a good idea to rename `factory.hpp` to `observables-factory.hpp` to avoid confusion with the factory used for data init.

lks1248 (approved):

PR #229, opened 2022-04-25
Fix focus transfer

sekelle:
This fixes a domain bug that occurs when an octree cell should be split and handed over to a different rank
in the same step. In order to a void the bug, we must first rebalance the tree part before handing it over to the new owner rank.


sekelle:
What version of clang-format is the CI using? Seems like it's screwing up here.

jgphpc (approved):
I will fix the format script in another pr

PR #230, opened 2022-04-27
VE fusion

jaescartin1:
VE added to the normal execution


sekelle (changes requested):
We will be able to use the individual parts of this PR, but we'll have to apply these changes incrementally:
1. introduce grad-h terms
2. modernize GPU particles data
3. change filenames to make VE the default for the CPU
4. introduce cuda ve implementation from this PR

For the moment I still need the old kernels for testing and debugging non-SPH parts of the code, but they can be made non-default to avoid interfering with work on the newer versions.

main/src/sphexa/propagator.hpp
```diff
@@ -172,132 +179,160 @@ class HydroProp final : public Propagator<DomainType, ParticleDataType>
         findNeighborsSfc<T, KeyType>(
             first, last, ngmax_, d.x, d.y, d.z, d.h, d.codes, d.neighbors, d.neighborsCount, domain.box());
         timer.step("FindNeighbors");
+#ifdef USE_CUDA
+        CHECK_CUDA_ERR(cudaMemcpy(d.devPtrs.d_codes, d.codes.data(), size_np_CodeType, cudaMemcpyHostToDevice));
+        timer.step("  * GPU CudaCopyIn Sync HostToDevice: codes");
+#endif
+
+#ifdef USE_CUDA
+        CHECK_CUDA_ERR(cudaMemcpy(d.devPtrs.d_x, d.x.data(), size_np_T, cudaMemcpyHostToDevice));
```
rmcabezon, line 185: As a suggestion to clean this a bit out in the future: these copies could be added in a single function (f.e. Host2Device) that takes the different fields as inputs. Then we can have a single line instruction while still keeping explicitly written the data that is being transferred.

main/src/sphexa/propagator.hpp
```diff
@@ -172,132 +179,160 @@ class HydroProp final : public Propagator<DomainType, ParticleDataType>
...
+        CHECK_CUDA_ERR(cudaMemcpy(d.devPtrs.d_z, d.z.data(), size_np_T, cudaMemcpyHostToDevice));
+        CHECK_CUDA_ERR(cudaMemcpy(d.devPtrs.d_h, d.h.data(), size_np_T, cudaMemcpyHostToDevice));
+        CHECK_CUDA_ERR(cudaMemcpy(d.devPtrs.d_m, d.m.data(), size_np_T, cudaMemcpyHostToDevice));
+        CHECK_CUDA_ERR(cudaMemcpy(d.devPtrs.d_vx, d.vx.data(), size_np_T, cudaMemcpyHostToDevice));
+        CHECK_CUDA_ERR(cudaMemcpy(d.devPtrs.d_vy, d.vy.data(), size_np_T, cudaMemcpyHostToDevice));
+        CHECK_CUDA_ERR(cudaMemcpy(d.devPtrs.d_vz, d.vz.data(), size_np_T, cudaMemcpyHostToDevice));
+        timer.step("  * GPU CudaCopyIn Sync HostToDevice: x,y,z, h, m, vx,vy,vz");
+#endif
+        computeRhoZero(first, last, ngmax_, d, domain.box());
+#ifdef USE_CUDA
```
rmcabezon, line 196: Suggestion for future upgrade: this structure of copy to host - MPI sync - copy back to the device, could also be included inside of the exchangeHalos function and clean it out from the propagator, as it appears many times.

main/src/sphexa/propagator.hpp
```diff
@@ -172,132 +179,160 @@ class HydroProp final : public Propagator<DomainType, ParticleDataType>
...
+        CHECK_CUDA_ERR(cudaMemcpy(d.devPtrs.d_alpha, d.alpha.data(), size_np_T, cudaMemcpyHostToDevice));
+        timer.step("  * GPU CudaCopyIn Sync HostToDevice: alpha");
+#endif
+
+        computeMomentumEnergy(first, last, ngmax_, d, domain.box());
+#ifdef USE_CUDA
+        CHECK_CUDA_ERR(cudaMemcpy(d.du.data(), d.devPtrs.d_du, size_np_T, cudaMemcpyDeviceToHost));
+
+        if (d.g == 0)
```
rmcabezon, line 283: Why do we use g==0 as condition here if we also have doGrav boolean?

main/src/sphexa/propagator.hpp
```diff
@@ -172,132 +179,160 @@ class HydroProp final : public Propagator<DomainType, ParticleDataType>
...
+        {
+            CHECK_CUDA_ERR(cudaMemcpy(d.grad_P_x.data(), d.devPtrs.d_grad_P_x, size_np_T, cudaMemcpyDeviceToHost));
+            CHECK_CUDA_ERR(cudaMemcpy(d.grad_P_y.data(), d.devPtrs.d_grad_P_y, size_np_T, cudaMemcpyDeviceToHost));
+            CHECK_CUDA_ERR(cudaMemcpy(d.grad_P_z.data(), d.devPtrs.d_grad_P_z, size_np_T, cudaMemcpyDeviceToHost));
+        }
```
rmcabezon, line 288: These copies to host are done, no matter we have gravity or not. They can be moved out of the if condition down to after gravity is calculated.

main/src/sphexa/propagator.hpp
```diff
@@ -172,132 +179,160 @@ class HydroProp final : public Propagator<DomainType, ParticleDataType>
...
+#endif
+        timer.step("MomentumEnergy");
 
         d.egrav = 0.0;
         if (doGrav)
         {
             mHolder_.upsweep(d, domain);
```
rmcabezon, line 295: We should add here a comment stating that gravity modifies gradP adding the gravitational forces.

main/src/sphexa/sphexa.cpp
```diff
@@ -136,11 +130,9 @@ int main(int argc, char** argv)
     // we want about 100 global nodes per rank to decompose the domain with +-1% accuracy
     size_t bucketSize = std::max(bucketSizeFocus, d.numParticlesGlobal / (100 * numRanks));
     Domain domain(rank, numRanks, bucketSize, bucketSizeFocus, theta, box);
-    auto   propagator = propagatorFactory<Domain, Dataset>(ve, ngmax, ng0, output, rank);
+    auto   propagator = propagatorFactory<Domain, Dataset>(ngmax, ng0, output, rank);
 
-    if (ve)
-        domain.sync(d.codes, d.x, d.y, d.z, d.h, d.m, d.u, d.vx, d.vy, d.vz, d.x_m1, d.y_m1, d.z_m1, d.du_m1, d.alpha);
-    else if (haveGrav)
+    if (haveGrav)
```
rmcabezon, line 132: Yet another boolean for gravity calculation (g=0, doGrav, haveGrav). We should normalize this.

main/src/sphexa/sphexa.cpp
```diff
@@ -171,7 +163,7 @@ int main(int argc, char** argv)
         if (d.iteration % 50 == 0) { viz::execute(d, domain.startIndex(), domain.endIndex()); }
     }
 
-    totalTimer.step("Total execution time of " + std::to_string(d.iteration - startIteration + 1) + " iterations of " +
+    totalTimer.step("Total execution time of " + std::to_string(d.iteration - startIteration) + " iterations of " +
```
rmcabezon, line 163: Shouldn't the +1 be here? When iteration = startIteration that would give zero instead of 1.

sph/include/sph/cuda/gpu_particle_data.cuh
```diff
@@ -31,28 +31,39 @@
...
+
 template<typename T, class KeyType>
 class DeviceParticlesData
 {
     size_t allocatedDeviceMemory = 0;
     size_t allocatedTaskSize     = 0;
 
 public:
-    // number of CUDA streams to use
+    // Number of CUDA streams to use
```
rmcabezon, line 63: Probably number of streams (NST) and taskSize should also go to CudaConfig.

sph/include/sph/cuda/gpu_particle_data.cuh
```diff
@@ -31,28 +31,39 @@
...
+#include "cuda_utils.cuh"
+
 namespace sphexa
 {
 namespace sph
 {
 namespace cuda
 {
 
+struct CudaConfig
```
rmcabezon, line 47: This CudaConfig should be somewhere else more exposed to the user.

rmcabezon (approved):
Very necessary homogenization of the names of the functions, elimination of the standard old-fashion SPH and implementation of cuda VE. I only added a few suggestions that can be addressed in subsequent PRs.

rmcabezon:
I agree with the steps suggested by Sebastian, but only if these don't take forever. The next test in line will need high-particle count and VE to work. I will do the gradh PR asap, so please keep this moving forward with high priority. Otherwise, we will be stuck in the tests and facing constant merge conflicts.

jaescartin1:
I am agree with Sebastian too. Please, feel free to apply the plan.

Like I am done with the VE fusion, because this PR works (adding the little changes for improve that both of you suggested for improve it), I will move to the implementation of the "Wind-cloud collision" test where we need VE and a minimum of 4 million of particles (here GPU execution is almost mandatory).

sekelle:
Addressed here: https://github.com/unibas-dmi-hpc/SPH-EXA/pull/265

PR #233, opened 2022-04-29
Gradh

rmcabezon:
Gradh added + unit test
whomega has been deleted as it is used locally now.


rmcabezon:
Last commit worked slower just because I was running in "debug" instead of in "release". Release version runs normally.

main/src/init/factory.hpp.orig
```diff
@@ -0,0 +1,87 @@
...
+        {
+            throw std::runtime_error("supplied value of --init " +
+                                     (testCase.empty() ? "[empty string]" : "(\"" + testCase + "\")") +
+                                     " is not an existing file and does not refer to a supported test case\n");
+        }
+    }
+>>>>>>> develop
+}
+
+} // namespace sphexa
```
sekelle: Is this file needed?

main/src/observables/factory.hpp.orig
```diff
@@ -0,0 +1,116 @@
...
+        H5PartCloseFile(h5_file);
+
+        if (attrValue) { return std::make_unique<TimeEnergyGrowth<Dataset>>(constantsFile); }
+    }
+
+    return std::make_unique<TimeAndEnergy<Dataset>>(constantsFile);
+>>>>>>> develop
+}
+
+} // namespace sphexa
```
sekelle: needed?

main/src/observables/time_energy_growth.hpp.orig
```diff
@@ -0,0 +1,172 @@
...
+
+        if (rank == 0)
+        {
+            fileutils::writeColumns(
+                constantsFile, ' ', d.iteration, d.ttot, d.minDt, d.etot, d.ecin, d.eint, d.egrav, khgr);
+        }
+    }
+};
+
+} // namespace sphexa
```
sekelle: needed?

main/src/init/factory.hpp.orig
rmcabezon: I don't think so. They were pulled from the last develop commit. I think they were created by Lukas and he forgot to remove.

main/src/observables/factory.hpp.orig
rmcabezon: Same as above.

main/src/observables/time_energy_growth.hpp.orig
rmcabezon: Same as above.

sekelle (approved):

PR #237, opened 2022-05-04
Return density_ve results by value

sekelle:


sph/include/sph/kernel_ve/density_kern.hpp
```diff
@@ -56,10 +55,10 @@ CUDA_DEVICE_HOST_FUN inline void densityJLoop(int i, T sincIndex, T K, const cst
     T wrho0i = wrho0[i];
     T xmassi = m[i] / rho0i;
 
-    T hInv  = 1.0 / hi;
+    T hInv  = T(1) / hi;
     T h3Inv = hInv * hInv * hInv;
 
-    T kxi      = 0.0;
+    T kxi      = xmassi;
```
rmcabezon, line 61: Add a comment here before these kxi initialization: //self-contribution initialization. We can have a similar structure in other Jloops.

sph/include/sph/kernel_ve/density_kern.hpp
```diff
@@ -56,10 +55,10 @@ CUDA_DEVICE_HOST_FUN inline void densityJLoop(int i, T sincIndex, T K, const cst
...
     T whomegai = 0.0;
```
rmcabezon, line 62: You can do the same with whomegai:
whomegai = - T(3) * xmassi

sph/include/sph/kernel_ve/density_kern.hpp
```diff
@@ -69,20 +68,24 @@ CUDA_DEVICE_HOST_FUN inline void densityJLoop(int i, T sincIndex, T K, const cst
...
 
-    kx[i]    = K * (kxi + xmassi) * h3Inv;
-    whomegai = K * (whomegai - 3.0 * xmassi) * h3Inv * hInv;
+    kxi *= K * h3Inv;
 
-    ro[i]    = kx[i] * m[i] / xmassi;
-    whomegai = whomegai * m[i] / xmassi + (ro[i] / rho0i - K * xmassi * h3Inv) * wrho0i;
-    T dhdrho = -hi / ro[i] / 3.0; // This /3 is the dimension hard-coded.
-    gradh[i] = 1.0 - dhdrho * whomegai;
+    whomegai = K * (whomegai - T(3) * xmassi) * h3Inv * hInv;
```
rmcabezon, line 80: If the self-contribution initialization for whomegai is done before the loop, you can change this to:
whomegai *= K * h3Inv * hInv

sph/include/sph/kernel_ve/density_kern.hpp
```diff
@@ -69,20 +68,24 @@ CUDA_DEVICE_HOST_FUN inline void densityJLoop(int i, T sincIndex, T K, const cst
...
+    whomegai = whomegai * rho0i + (kxi - K * xmassi * h3Inv) * wrho0i;
+
+    T dhdrho = -hi / (kxi * rho0i * T(3)); // This /3 is the dimension hard-coded.
+
+    T rhoi   = kxi * rho0i;
```
rmcabezon, line 85: VEs are hardcoded here. It is better to keep the flexibility to be able to change the structure of the VEs in xmass.

rmcabezon (commented):
The exploit of rho = kx * rho0 hardcodes the shape of the VEs. If we can save memory somewhere else would be better to keep the flexibility to simply change Xmass. Otherwise, it will be very cumbersome and prone to error to change all the equations later.

sph/include/sph/kernel_ve/density_kern.hpp
sekelle: Will do and move to new PR.

sph/include/sph/kernel_ve/density_kern.hpp
sekelle: same as above

sph/include/sph/kernel_ve/density_kern.hpp
sekelle: nice simplification

sph/include/sph/kernel_ve/density_kern.hpp
sekelle: will be addressed in new PR

PR #238, opened 2022-05-04
Mom ener simplify

sekelle:
Storage of the field "rho" is avoided as we can compute rho by multiplying kx and rho0

### DIFF
--- a/sph/include/sph/kernel_ve/momentum_energy_kern.hpp
+++ b/sph/include/sph/kernel_ve/momentum_energy_kern.hpp
@@ -46,10 +46,10 @@ template<typename T>
 CUDA_DEVICE_HOST_FUN inline void
 momentumAndEnergyJLoop(int i, T sincIndex, T K, const cstone::Box<T>& box, const int* neighbors, int neighborsCount,
                        const T* x, const T* y, const T* z, const T* vx, const T* vy, const T* vz, const T* h,
-                       const T* m, const T* ro, const T* p, const T* c, const T* c11, const T* c12, const T* c13,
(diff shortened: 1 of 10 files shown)


sph/include/sph/kernel_ve/momentum_energy_kern.hpp
```diff
@@ -192,16 +194,16 @@ momentumAndEnergyJLoop(int i, T sincIndex, T K, const cstone::Box<T>& box, const
...
             //           vz_ij * (a * termA3_i + b * termA3_j);
         }
     }
 
     // with the choice of calculating coordinate (r) and velocity (v_ij) differences as i - j,
     // we add the negative sign only here at the end instead of to termA123_ij in each iteration
-    a_visc_energy = std::max(T(0), a_visc_energy);
-    du[i]         = T(0.5) * (energy + a_visc_energy);
+    a_visc_energy = (a_visc_energy > 0) ? a_visc_energy : T(0);
+    du[i]         = energy + T(0.5) * a_visc_energy;
```
rmcabezon, line 206: Add a comment stating that the 2 from 2P/rho has been simplified here, or we might add it by mistake in the future.

rmcabezon (commented):
Same comments as in the other PR regarding the usage of rho0 to avoid storing rho.

sph/include/sph/kernel_ve/momentum_energy_kern.hpp
sekelle: will do and transfer to the new PR

PR #241, opened 2022-05-05
Explicit VE definition

sekelle:
xmass = m / rhoZero is a particular choice of defining generalized volume elements.
We make this explicit in the code such that this choice can later be revisited and changed if needed.


sph/include/sph/iad.hpp
```diff
@@ -45,26 +45,26 @@ void computeIADImpl(size_t startIndex, size_t endIndex, size_t ngmax, Dataset& d
...
-                          ro,
-                          wh,
-                          whd,
-                          c11,
-                          c12,
-                          c13,
-                          c22,
-                          c23,
-                          c33);
+        kernels::IADJLoop3L(i,
```
sekelle, line 48: I've started to use the "plain" names for the VE functions and instead add the suffix "3L" to the old kernels
where needed to avoid name clashes.

sph/include/sph/kernel_ve/av_switches_kern.hpp
```diff
@@ -98,8 +98,8 @@ AVswitchesJLoop(int i, T sincIndex, T K, const cstone::Box<T>& box, const int* n
         T rv           = rx * vx_ij + ry * vy_ij + rz * vz_ij;
         T vijsignal_ij = 0.0;
 
-        if (rv < 0.0) { vijsignal_ij = ci + c[j] - 3.0 * rv / dist; }
-        vijsignal_i = std::max(vijsignal_i, vijsignal_ij);
+        if (rv < T(0)) { vijsignal_ij = ci + c[j] - T(3) * rv / dist; }
+        vijsignal_i = stl::max(vijsignal_i, vijsignal_ij);
```
rmcabezon, line 102: Why the use of STL instead of STD here?

sph/include/sph/kernel_ve/av_switches_kern.hpp
```diff
@@ -108,21 +108,21 @@ AVswitchesJLoop(int i, T sincIndex, T K, const cstone::Box<T>& box, const int* n
...
-    T graddivv = sqrt(graddivv_x * graddivv_x + graddivv_y * graddivv_y + graddivv_z * graddivv_z);
+    T graddivv = std::sqrt(graddivv_x * graddivv_x + graddivv_y * graddivv_y + graddivv_z * graddivv_z);
 
     T alphaloc = 0.0;
-    if (divv_i < 0.0)
+    if (divv_i < T(0))
     {
         T a_const = hi * hi * graddivv;
-        alphaloc  = alphamax * a_const / (a_const + hi * abs(divv_i) + 0.05 * ci);
+        alphaloc  = alphamax * a_const / (a_const + hi * abs(divv_i) + T(0.05) * ci);
```
rmcabezon: Should this be std::abs ?

rmcabezon (approved):

sph/include/sph/kernel_ve/av_switches_kern.hpp
sekelle: `std::max` is not supported on the GPU, that's why we have this as an alternative. It is identical to the `std` version, but has the GPU function qualifier.

sph/include/sph/kernel_ve/av_switches_kern.hpp
sekelle: Yes, I'll change that.

PR #242, opened 2022-05-06
Compute rho on-the-fly

sekelle:
With volume elements, we can compute rho = kx * m / xmass where needed in momentum_and_energy to save memory.
Since kx, m and xmass are already used and loaded at that place, we also save memory-bandwidth and reduce the cache footprint.

Currently, this means that in order to make plots involving rho, kx, m and xmass have to be selected for output instead.
But I'll fix this in a follow-up PR, such that if rho is selected for output, it will be reconstructed from kx, m and xmass
prior to writing the output.


sph/test/kernel_ve/momentum_energy_kern.cpp
```diff
@@ -142,113 +140,9 @@ TEST(MomentumEnergy, JLoop)
                                          &du,
                                          &maxvsignal);
 
-    EXPECT_NEAR(grad_Px, -4.7423317200544224e-1, 1e-10);
-    EXPECT_NEAR(grad_Py, 8.737996639917453e-2, 1e-10);
-    EXPECT_NEAR(grad_Pz, -5.303834041425971e-1, 1e-10);
-    EXPECT_NEAR(du, -3.8485025262537881e-3, 1e-10);
```
sekelle, line 148: The reference changes here slightly, because the supplied input rho was not consistent with rho = kx * m / xmass.

PR #243, opened 2022-05-06
Compute wrho0 together with whomega

sekelle:
Saves another field and also gives a little speed-up.
This is based on the PR about computing rho on the fly. It should be merged before reviewing this, such that the
diff is minimized for easier reviewing.


sph/test/kernel_ve/density_kern.cpp
```diff
@@ -88,12 +86,10 @@ TEST(Density, JLoop)
                                wh.data(),
                                whd.data(),
                                xm.data(),
-                               wrho0.data(),
-                               rho.data(),
                                kx.data(),
                                gradh.data());
-    EXPECT_NEAR(rho[0], 1.67849454056818e-2, 1e-10);
-    EXPECT_NEAR(gradh[0], 1.2501347388453987, 1e-10);
```
sekelle, line 94: The gradh reference had to be changed because the input whro0 was inconsistent with the whro0 that is computed on the fly.
Before changing it, I added a runtime check that the previous whro0 computed in Xmass is identical to the whro0 computed on-the-fly in density.

sph/include/sph/eos.hpp
```diff
@@ -18,33 +18,77 @@ namespace sph
  *
  * In this simple version of state equation, we calculate all depended quantities
  * also for halos, not just assigned particles in [startIndex:endIndex], so that
- * we can avoid halo exchange of p and c.
+ * we could potentially avoid halo exchange of p and c in return for exchanging halos of u.
```
rmcabezon: That's true probably only for the ideal gas EOS. But we are going to change it for a more complex one where there is no simple relation between these magnitudes.

sph/include/sph/eos.hpp
```diff
@@ -18,33 +18,77 @@ namespace sph
...
-        T tmp = u[i] * (gamma - 1);
+        T tmp = u[i] * (gamma - T(1));
+        T rho = kx[i] * m[i] / xm[i];
+        p[i]  = rho * tmp;
+        c[i]  = std::sqrt(tmp);
+    }
+}
+
+/*! @brief Ideal gas EOS for SPH formulations where rho is stored
+ *
```
rmcabezon: Is this version saved temporarily for checking?

sph/include/sph/kernel_ve/density_kern.hpp
```diff
@@ -61,6 +60,7 @@ CUDA_DEVICE_HOST_FUN inline void densityJLoop(int i, T sincIndex, T K, const cst
     // initialize with self-contribution
     T kxi      = xmassi;
     T whomegai = -T(3) * xmassi;
+    T wrho0i   = -T(3) * mi;
```
rmcabezon, line 63: Nice optimization to calculate wrho0 here.

rmcabezon (approved):

sph/include/sph/eos.hpp
sekelle: yes

sph/include/sph/eos.hpp
sekelle: Yes I thought that this is the case and refrained from implementing this...

PR #245, opened 2022-05-06
Fuse particle loops of IAD and divV-curlV

sekelle:
Improves cache locality a bit.
For the GPU, this allows to launch a single kernel for those two steps, also resulting in better cache locality
and easier porting to CUDA.


rmcabezon (commented):
What's the objective of this optimization? I don't see what is to gain apart from losing modularity.

Ah! Ok. I didn't see your comment. Not superfan of mixing two different calculations in the same module, but if there is a clear gain, it is ok. We anyway need both all the time.

rmcabezon (approved):

sekelle:
I understand the concern. On the other hand, the actual j-loop functions for IAD and divv-curlv are still separate and are also still tested separately. If the need arises, we can quickly launch those two operations separately again.

PR #247, opened 2022-05-09
Propagator sync

sekelle:
- Adds wrapper for domain::sync to propagator for reuse in initialization
- Allow VE + gravity

### DIFF
--- a/main/src/sphexa/propagator.hpp
+++ b/main/src/sphexa/propagator.hpp
@@ -130,9 +130,9 @@ template<class DomainType, class ParticleDataType>
 class HydroProp final : public Propagator<DomainType, ParticleDataType>
 {
     using Base = Propagator<DomainType, ParticleDataType>;
+    using Base::doGravity_;
     using Base::ng0_;
     using Base::ngmax_;
-    using Base::doGravity_;
     using Base::timer;
 
     using T             = typename ParticleDataType::RealType;
@@ -227,9 +227,9 @@ template<class DomainType, class ParticleDataType>
 class HydroVeProp final : public Propagator<DomainType, ParticleDataType>
 {
     using Base = Propagator<DomainType, ParticleDataType>;
+    using Base::doGravity_;
     using Base::ng0_;
     using Base::ngmax_;
-    using Base::doGravity_;
     using Base::timer;
 
     using T             = typename ParticleDataType::RealType;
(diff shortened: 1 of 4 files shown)


main/src/sphexa/propagator.hpp
```diff
@@ -73,6 +76,8 @@ class Propagator
     //! average number of neighbors per particle
     size_t ng0_;
```
rmcabezon, line 77: The average number of neighbors per particle is a very useless magnitude. Tracking the number of neighbors on each particle is done because if there is an issue on a working code with a specific scenario, it might be due to one (or few) particles that get super high or super low amounts of neighbors. This will be basically unseeable on the average count. Why do we keep this magnitude? Is it used somewhere?

main/src/sphexa/propagator.hpp
```diff
@@ -270,6 +301,18 @@ class HydroVeProp final : public Propagator<DomainType, ParticleDataType>
...
         timer.step("MomentumAndEnergy");
+
+        d.egrav = 0.0;
+        if (doGravity_)
+        {
+            mHolder_.upsweep(d, domain);
+            timer.step("Upsweep");
+            mHolder_.traverse(d, domain);
+            // temporary sign fix, see note in ParticlesData
+            d.egrav = (d.g > 0.0) ? d.egrav : -d.egrav;
```
rmcabezon, line 312: Why do we need this check here? G is always positive. Either 1 or 6.67e-8.

rmcabezon (approved):
Just a couple of comments.

main/src/sphexa/propagator.hpp
sekelle: This is a misnomer on my part, this should be called "target number of neighbors per particles". The smoothing lengths
are adapted each step to aim at this number.

main/src/sphexa/propagator.hpp
sekelle: Yes indeed. The gravity module reports acceleration, whereas momentum_energy reports negative acceleration.
This negative-G hack is used to reconcile that discrepancy. IMHO, the simplest solution would be to make momentum_energy j-loop return grad_P by value (instead of direct memory access with a pointer), such that we can keep the usual convention of returning gradient of pressure for the j-loop function. The calling function, that does the i-loop can then decide to store grad_P back to memory as acceleration with the negative sign.

PR #250, opened 2022-05-09
Restore rho output

sekelle:
This allows to output "rho" as usual also with volume elements, even though it is not an active field.
Whenever an output is triggered, the implementation computes rho as rho = kx * m / xmass.

### DIFF
--- a/sph/include/sph/particles_data.hpp
+++ b/sph/include/sph/particles_data.hpp
@@ -98,7 +98,7 @@ public:
     std::vector<T>       dt, dt_m1;                    // timestep
     std::vector<T>       c11, c12, c13, c22, c23, c33; // IAD components
     std::vector<T>       alpha;                        // AV coeficient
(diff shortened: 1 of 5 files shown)


sph/include/sph/particles_data.hpp
```diff
@@ -102,7 +99,6 @@ class ParticlesData
     std::vector<T>       c11, c12, c13, c22, c23, c33; // IAD components
     std::vector<T>       alpha;                        // AV coeficient
     std::vector<T>       xm;                           // Classical SPH density
```
rmcabezon: This is "volume element definition" instead of "Classical SPH density"

rmcabezon (approved):
I would use the same structure to create another derived field for the radius, as it is used in the analysis as much as density, if not more.

sph/include/sph/particles_data.hpp
sekelle: ack

PR #251, opened 2022-05-10
added Gravitational Waves observable

lks1248:


main/src/observables/gravitational_waves.hpp
```diff
@@ -0,0 +1,195 @@
...
+
+#include "iobservables.hpp"
+#include "io/ifile_writer.hpp"
+
+namespace sphexa
+{
+
+//!@brief calculates the second derivative of the quadpole momentum
+template<class T, class Dataset>
+T d2QuadpoleMomentum(size_t begin, size_t end, int pos1, int pos2, Dataset& d)
```
sekelle: This should only take the arrays that it needs listed below, rather than the complete Dataset.
grad_P should actually be called acceleration (ax, ay, az) here and you will have to change the sign once https://github.com/unibas-dmi-hpc/SPH-EXA/pull/252 is merged

sekelle: dim1 and dim2 would be better names for pos1 and pos2

main/src/observables/gravitational_waves.hpp
```diff
@@ -0,0 +1,195 @@
...
+    T* vy       = d.vy.data();
+    T* vz       = d.vz.data();
+    T* gradP_x  = d.grad_P_x.data();
+    T* gradP_y  = d.grad_P_y.data();
+    T* gradP_z  = d.grad_P_z.data();
+    T* m        = d.m.data();
+
+    T out = 0.0;
+
+    std::array<T*, 3> pos = {x, y, z};
```
sekelle: rename `pos` to `coords` for example

main/src/observables/gravitational_waves.hpp
sekelle: the argument arrays, x, v, accel should be passed as `const`

main/src/observables/gravitational_waves.hpp
```diff
@@ -0,0 +1,195 @@
...
+    std::array<T*, 3> gradP = {gradP_x, gradP_y, gradP_z};
+
+    if(pos1 == pos2)
+    {
+
+#pragma omp parallel for reduction(+ : out)
+        for(size_t i = begin; i < end; i++)
+        {
+            T acc1 = -gradP[pos1][i];
+            T factor1 = vx[i] * vx[i] + vy[i] * vy[i] + vz[i] * vz[i];
```
sekelle: since this is the square of the norm of the velocity, why not name it like that? e.g. `v2`?

main/src/observables/gravitational_waves.hpp
```diff
@@ -0,0 +1,195 @@
...
+            T factor2 = x[i] * (-gradP_x[i]) + y[i] * (-gradP_y[i]) + z[i] * (-gradP_z[i]);
```
sekelle: `coordDotAccel`?

main/src/observables/gravitational_waves.hpp
```diff
@@ -0,0 +1,195 @@
...
+    d2Local[3] = d2QuadpoleMomentum<T>(first, last, 0, 1, d); //ixy
+    d2Local[4] = d2QuadpoleMomentum<T>(first, last, 0, 2, d); //ixz
+    d2Local[5] = d2QuadpoleMomentum<T>(first, last, 1, 2, d); //iyz
+
+
+    int rootRank = 0;
+    std::array<T, 6> d2Global;
+    MPI_Reduce(d2Local.data(), d2Global.data(),6, MpiType<T>{}, MPI_SUM, rootRank, MPI_COMM_WORLD);
+
+    if (viewTheta == 0.0 && viewPhi == 0.0)
```
sekelle: 1. Is this `if` really needed, doesn't the case below for theta, phi != 0 also work if you pass theta,phi == 0? Note how this opens a coverage gap in the test that you implemented: the case theta, phi == 0 is not tested.
2. The calculation of `httplus, httcross` from d2Global should be its own function with a unit test.

main/test/CMakeLists.txt
```diff
@@ -3,12 +3,13 @@ set(UNIT_TESTS
...
         test_main.cpp)
 
 set(exename frontend_units)
 add_executable(${exename} ${UNIT_TESTS})
 target_compile_options(${exename} PRIVATE -Wall -Wextra -Wno-unknown-pragmas)
 
-target_include_directories(${exename} PRIVATE ${CSTONE_DIR} ${PROJECT_SOURCE_DIR}/main/src)
-target_link_libraries(${exename} PRIVATE GTest::gtest_main)
+target_include_directories(${exename} PRIVATE ${CSTONE_DIR} ${SPH_DIR} ${PROJECT_SOURCE_DIR}/main/src)
+target_link_libraries(${exename} PRIVATE GTest::gtest_main ${MPI_CXX_LIBRARIES})
```
sekelle: We don't enable MPI in unit tests because multi-rank programs are harder to debug, and, on HPC systems can't be run on login nodes. If you really need MPI, you need to create a standalone test executable and
in that case, you're not really writing a unit test anymore. You're combining multiple units and at that point what you're writing is an integration test.

What you need is a unit test for `d2QuadpoleMomentum` and another one for the `httplus-cross` function that you're going to create. This doesn't need the dependency on ParticlesData and doesn't need MPI and would give you full code coverage except for the reduction call. It's probably not needed, but if you then still want, you can write an MPI-aware integration test that actually tests the MPI reduction. This is something that your test does not do, if you execute it on more than one rank, it will fail.

sekelle (changes requested):

rmcabezon:
Please change d2Global and d2Local to more meaningful names, such as d2Qxx_local, d2Qxy_global, and so.
For people working on GWs d2Qxx is well understood.

main/src/observables/grav_waves_calculations.hpp
```diff
@@ -0,0 +1,103 @@
...
+ *  @author Lukas Schmidt
+ */
+
+#include <cmath>
+
+namespace sphexa
+{
+
+template<class T>
+void computeHtt(std::array<T, 6> quadpoleMomentum, T theta, T phi, T* httplus, T* httcross)
```
sekelle, line 39: This needs a docstring.

main/src/observables/gravitational_waves.hpp
```diff
@@ -0,0 +1,125 @@
...
+ * @param last          first locally assigned particle index of buffers in @p d
+ * @param viewTheta     viewing angle to determine ...
+ * @param viewPhi       viewing angle to determine ...
+ * @return              array containing the observables and the second derivative of the quadpole momentum:
+ *                      {httplus, httcross, ixx, iyy, izz, ixy, ixz, iyz}
+ */
+template<class T, class Dataset>
+std::array<T, 8> gravRad(Dataset& d,size_t first, size_t last, T viewTheta, T viewPhi)
+{
+
```
sekelle, line 67: Introduce this here:
```c++

struct Dim
{
enum IndexLabels
{
    x = 0,
    y = 1,
    z = 2
};
};

struct Q
{
enum IndexLabels
{
    xx = 0,
    xy = 1,
    xz = 2,
    ...
};
};

File: main/src/observables/gravitational_waves.hpp [hunk 9]
```diff
@@ -0,0 +1,125 @@
...
+    std::array<T, 6> d2Qxx_local;
```
sekelle: rename to `d2Q_local`

main/src/observables/gravitational_waves.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+    d2Qxx_local[0] = d2QuadpoleMomentum<T>(first, last, 0, 0,
```
sekelle: These 6 calls now become
```c++
d2Q_local[Q::xx] = d2QuadpoleMomentum<T>(first, last, Dim::x, Dim::x,
```
etc

sekelle: This should address Rubens suggestion about giving understandable names, while
keeping the 6 values in an array which is needed for the global reduction.

sekelle:
Needs clang-format.

sekelle:
cscs-ci run

sekelle:
I'll merge this as soon as we have CI passing.

lks1248:
great, thanks!

jgphpc:
cscs-ci run

sekelle (approved):

PR #252, opened 2022-05-12
Fix negative G

sekelle:
Unify sign conventions of grad_P to harmonize with the integrator and gravity.

### DIFF
--- a/sph/include/sph/eos.hpp
+++ b/sph/include/sph/eos.hpp
@@ -15,8 +15,8 @@ CUDA_DEVICE_HOST_FUN util::tuple<T2, T2> equationOfState(T1 u, T2 rho)
     constexpr T2 gamma = (5.0 / 3.0);
 
     T2 tmp = u * (gamma - T1(1));
-    T2 p  = rho * tmp;
-    T2 c  = std::sqrt(tmp);
+    T2 p   = rho * tmp;
+    T2 c   = std::sqrt(tmp);
 
     return {p, c};
 }
@@ -61,7 +61,7 @@ void computeEquationOfState(size_t startIndex, size_t endIndex, Dataset& d)
 #pragma omp parallel for schedule(static)
     for (size_t i = startIndex; i < endIndex; ++i)
     {
-        auto rho = kx[i] * m[i] / xm[i];
+        auto rho             = kx[i] * m[i] / xm[i];
         std::tie(p[i], c[i]) = equationOfState(u[i], rho);
     }
 }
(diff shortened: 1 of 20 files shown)


sph/include/sph/eos.hpp
```diff
@@ -9,6 +9,33 @@ namespace sphexa
...
+
+    T2 cv   = T2(1.5) * R / mui;
+    T2 temp = u / cv;
+    T2 tmp  = u * (gamma - T1(1));
+    T2 p    = rho * tmp;
+    T2 c    = std::sqrt(tmp);
+
+    return {p, c, cv, temp};
+}
+
```
rmcabezon, line 38: These two EOS are named the same. How we differentiate between them at the time of choosing one?

sph/include/sph/kernel/momentum_energy_kern.hpp
```diff
@@ -138,9 +138,9 @@ momentumAndEnergyJLoop(int i, T sincIndex, T K, const cstone::Box<T>& box, const
     // with the choice of calculating coordinate (r) and velocity (v_ij) differences as i - j,
     // we add the negative sign only here at the end instead of to termA123_ij in each iteration
     du[i]       = -K * T(0.5) * energy;
-    grad_P_x[i] = -K * momentum_x;
-    grad_P_y[i] = -K * momentum_y;
-    grad_P_z[i] = -K * momentum_z;
+    grad_P_x[i] = K * momentum_x;
+    grad_P_y[i] = K * momentum_y;
+    grad_P_z[i] = K * momentum_z;
     *maxvsignal = maxvsignali;
```
rmcabezon, line 144: This sign was not related to the gravity, but to the choice of the way of calculating distances for termA1_i, termA2_i, and so... Does removing it still takes this into account?

rmcabezon: It is also opposite sign compared to the VE implementation of momentum.

rmcabezon (approved):

sph/include/sph/eos.hpp
sekelle: They are the same EOS, but one computes additional quantities. The differentiation occurs by the signature, i.e. different number of arguments.

sph/include/sph/kernel/momentum_energy_kern.hpp
sekelle: The sign is still correct. It's opposite to the VE implementation because of opposite signs of the A-terms.

PR #253, opened 2022-05-12
Add fftw dependency

jaescartin1:


jaescartin1:
You can find the source module in: https://github.com/egpbos/findFFTW

CMakeLists.txt
```diff
@@ -103,6 +103,8 @@ elseif(INSITU STREQUAL "Ascent")
 endif()
 # }}}
 
+find_package(FFTW)
```
sekelle: Fine in principle, but I think FFTW should be optional, like we handle H5Part for instance.

jaescartin1: Added as optional FFTW

sekelle (approved):

PR #254, opened 2022-05-14
Field states

sekelle:
Introduction of field states: unallocated fields can be allocated ("acquired") at the expense of another field ("released") at zero cost. This PR only adds the capability, it is not used yet in any propagator.

### DIFF
--- a/main/src/init/file_init.hpp
+++ b/main/src/init/file_init.hpp
@@ -89,7 +89,7 @@ public:
         cstone::Box<T> box(
             extents[0], extents[1], extents[2], extents[3], extents[4], extents[5], pbc[0], pbc[1], pbc[2]);
 
-        resize(d, count);
+        d.resize(count);
 
         H5PartSetView(h5_file, first, last - 1);
         h5part_int64_t errors = H5PART_SUCCESS;
--- a/main/src/init/file_init.hpp
+++ b/main/src/init/file_init.hpp
@@ -106,7 +106,6 @@ public:
         initField(h5_file, rank, d.vy, "vy", 0.0);
         initField(h5_file, rank, d.vz, "vz", 0.0);
 
-        initField(h5_file, rank, d.dt_m1, "dt_m1", d.minDt);
         initField(h5_file, rank, d.du_m1, "du_m1", 0.0);
         initField(h5_file, rank, d.alpha, "alpha", d.alphamin);
 
--- a/main/src/observables/gravitational_waves.hpp
+++ b/main/src/observables/gravitational_waves.hpp
@@ -1,0 +1,233 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2021 CSCS, ETH Zurich
+ *               2021 University of Basel
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+/*! @file output time, energies and the gravitational radiation as well as the quadrupole moment
+ *
+ *  @author Lukas Schmidt
+ */
+
+#include <array>
+#include "mpi.h"
+#include "iobservables.hpp"
+#include "io/ifile_writer.hpp"
+#include "grav_waves_calculations.hpp"
+
+namespace sphexa
+{
+
+/*!@brief calculates the gravitational waves observable
+ *
+ * @tparam T            floating point type
+ * @tparam Dataset
+ * @param d             Dataset
+ * @param first         first locally assigned particle index of buffers in @p d
+ * @param last          first locally assigned particle index of buffers in @p d
+ * @param viewTheta     viewing angle for the polarization modes
+ * @param viewPhi       viewing angle for the polarization modes
+ * @return              array containing the polarization modes and the second derivative of the quadpole momentum:
+ *                      {httplus, httcross, ixx, iyy, izz, ixy, ixz, iyz}
+ */
+template<class T, class Dataset>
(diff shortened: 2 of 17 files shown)


main/src/init/file_init.hpp
```diff
@@ -106,7 +106,6 @@ class FileInit : public ISimInitializer<Dataset>
         initField(h5_file, rank, d.vy, "vy", 0.0);
         initField(h5_file, rank, d.vz, "vz", 0.0);
 
-        initField(h5_file, rank, d.dt_m1, "dt_m1", d.minDt);
         initField(h5_file, rank, d.du_m1, "du_m1", 0.0);
```
rmcabezon, line 109: Why do we need to store the _m1 fields in the HDF5? Usually, it is enough to initialize them with the same value as the current states and use a small timestep.

main/src/observables/gravitational_waves.hpp
```diff
@@ -77,69 +77,157 @@ std::array<T, 8> gravRad(Dataset& d,size_t first, size_t last, T viewTheta, T vi
...
+                                             d.vz.data(),
+                                             d.ax.data(),
+                                             d.ay.data(),
+                                             d.az.data(),
+                                             d.m.data());
+
+    int              rootRank = 0;
     std::array<T, 6> d2Qxx_global;
-    MPI_Reduce(d2Q_local.data(), d2Qxx_global.data(),6, MpiType<T>{}, MPI_SUM, rootRank, MPI_COMM_WORLD);
+    MPI_Reduce(d2Q_local.data(), d2Qxx_global.data(), 6, MpiType<T>{}, MPI_SUM, rootRank, MPI_COMM_WORLD);
```
rmcabezon, line 167: Why is the global called d2Qxx? It would less confusing simply d2Q_global

main/src/observables/gravitational_waves.hpp
```diff
@@ -77,69 +77,157 @@ std::array<T, 8> gravRad(Dataset& d,size_t first, size_t last, T viewTheta, T vi
...
-    return {httplus, httcross, d2Qxx_global[0], d2Qxx_global[1], d2Qxx_global[2], d2Qxx_global[3], d2Qxx_global[4], d2Qxx_global[5]};
+    return {httplus,
+            httcross,
+            d2Qxx_global[0],
+            d2Qxx_global[1],
+            d2Qxx_global[2],
+            d2Qxx_global[3],
+            d2Qxx_global[4],
+            d2Qxx_global[5]};
 }
```
rmcabezon, line 182: This is very difficult to follow. d2Q_global should be implemented in a similar way as d2Q_local, where we can specify dimension coordinates, and no array indexes.

main/src/observables/gravitational_waves.hpp
```diff
@@ -77,69 +77,157 @@ std::array<T, 8> gravRad(Dataset& d,size_t first, size_t last, T viewTheta, T vi
...
+                                    d.eint,
+                                    d.egrav,
+                                    out[0],
+                                    out[1],
+                                    out[2],
+                                    out[3],
+                                    out[4],
+                                    out[5],
+                                    out[6],
+                                    out[7]);
```
rmcabezon, line 228: Same here. Out is a meaningless variable. It should have a name that has meaning to understand what is returning and where.

main/src/sphexa/propagator.hpp
```diff
@@ -244,14 +247,38 @@ class HydroVeProp final : public Propagator<DomainType, ParticleDataType>
...
+                       "curlv",
+                       "gradh",
+                       "keys",
+                       "nc");
+    }
+
     void sync(DomainType& domain, ParticleDataType& d) override
     {
-        if (doGravity_)
+        if (d.g != 0.0)
```
rmcabezon, line 281: Why do we still use a Universal Constant as a parameter, instead of a simple boolean that can be defined in the init, an input argument, or an input parameter file?

main/src/sphexa/propagator.hpp
```diff
@@ -323,36 +350,43 @@ class HydroVeProp final : public Propagator<DomainType, ParticleDataType>
...
+    {
+        bool outputRho =
+            std::find(d.outputFieldNames.begin(), d.outputFieldNames.end(), "rho") != d.outputFieldNames.end();
+        if (outputRho)
+        {
+            d.release("rho");
+            d.acquire("c11");
+        }
+    }
 };
```
rmcabezon, line 382: Please, use this same structure to allow the output of the radius as a field.

main/test/sphexa/particles_data.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    EXPECT_EQ(d.du.size(), 0);
+    EXPECT_EQ(d.c.size(), size);
+}
+
+TEST(ParticlesData, releaseAcquireInt)
+{
+    ParticlesData<double, unsigned, cstone::CpuTag> d;
+
+    d.setDependent("x");
```
rmcabezon, line 77: Can you set a dependent field that is named like a conserved field? Why is this setDependent("x") working?

main/test/sphexa/particles_data.cpp
```diff
@@ -0,0 +1,153 @@
...
+    int indexOfX = std::find(d.fieldNames.begin(), d.fieldNames.end(), "x") - d.fieldNames.begin();
+
+    size_t size = 10;
+    d.resize(size);
+
+    EXPECT_EQ(d.x.size(), size);
+    EXPECT_EQ(d.y.size(), 0);
+
+    d.release(indexOfX);
+    d.acquire("y");
```
rmcabezon, line 88: Is "y" not defined as a conserved field by default in SPH-EXA?
Does "y" inherit the integer nature of indexOfX?
Why there is no type mismatch here?

main/test/sphexa/particles_data.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    size_t size = 10;
+    d.resize(size);
+
+    EXPECT_EQ(d.du.size(), size);
+    EXPECT_EQ(d.ax.size(), size);
+    EXPECT_EQ(d.p.size(), size);
+
+    d.release("du", "ax", "p");
+    d.acquire("c11", "c12", "c13");
```
rmcabezon, line 108: Is the acquire going to be done in the same order as the release? I mean c11 will be allocated in the former du memory space?

main/src/observables/gravitational_waves.hpp
sekelle: This should be already done / fixed, might be left-over due to missing merge.

main/src/observables/gravitational_waves.hpp
sekelle: same

main/test/sphexa/particles_data.cpp
sekelle: No, this is not possible. `setDependent(x)` is only working because `x` was not declared dependent.

main/test/sphexa/particles_data.cpp
sekelle: "y" is declared as conserved in SPH-EXA. But this is not SPH-EXA, just an isolated unit test for a data container.
The integer here does not refer to the contents of "y" or "x", this is a demonstration that for release/acquire, fields may be
referred to either by string names or by the index into the table that defines the names.

main/test/sphexa/particles_data.cpp
sekelle: No, the implementation does not make any guarantees about the ordering. Also not possible since "du" and "c11" might be
a different type for example.

main/src/sphexa/propagator.hpp
sekelle: It's not really an immutable constant because it depends on the test case. Since we have to allow this constant to be modified by the user / the ICs, the current setup is just for convenience to only have to set one value (G) instead of two (G and the boolean) in the input. We could add the boolean as well but would also have to save that to the checkpoint file in order for restarts to work correctly.

main/src/init/file_init.hpp
sekelle: You don't need to. You can choose not to store them and reset the timestep to something small.
If you want to restore your system to the exact same state, then they are needed.

rmcabezon (approved):

PR #256, opened 2022-05-16
Added artificial conductivity

rmcabezon:
This adds artificial conductivity to the momentum equation.
The new function is implemented in kernels.hpp. There, you'll find a few lines commented that should be active when there is gravity and the other way around when there is not.  I leave it in this way for you to think of a proper way to implement this within a template, instead of using an if-condition or a pre-compiler directive.


sekelle (approved):
Looks good to me, I'll see about implementing the two version depending on whether gravity is active or not.

PR #261, opened 2022-05-18
added polytropic EOS

rmcabezon:
Polytropic EOS for cold 1.4 M_sun neutron stars added. Kpol is for now hardcoded to that specific mass of NS.
I'm not sure if I used the correct declarations as I don't know what T1, T2, and T3 are. So I used T2 for everything as it seems the default for the calculations.


sph/include/sph/eos.hpp
```diff
@@ -9,41 +9,80 @@ namespace sphexa
...
+ *
+ * @param u    internal energy
+ * @param rho  baryonic density
+ *
+ * This EOS is used for simple cases where we don't need the temperature.
+ * Returns pressure, speed of sound
+ */
 template<class T1, class T2>
-CUDA_DEVICE_HOST_FUN util::tuple<T2, T2> equationOfState(T1 u, T2 rho)
+CUDA_DEVICE_HOST_FUN auto equationOfState(T1 u, T2 rho)
```
sekelle: maybe `idealGasEOS` and `polytropicEOS` would be cleaner names?

sekelle (approved):

sph/include/sph/eos.hpp
rmcabezon: Sure! I will do this change.

sekelle:
cscs-ci run

PR #265, opened 2022-05-20
VE CUDA implementation

sekelle:


sph/include/sph/hydro_std/density_gpu.cu
```diff
@@ -91,9 +91,9 @@ void computeDensity(size_t startIndex, size_t endIndex, size_t ngmax, Dataset& d
...
         int* d_neighborsCount_use = d.devData.d_stream[sIdx].d_neighborsCount;
 
-        unsigned firstParticle       = startIndex + i * taskSize;
-        unsigned lastParticle        = std::min(startIndex + (i + 1) * taskSize, endIndex);
-        unsigned numParticlesCompute = lastParticle - firstParticle;
+        size_t firstParticle       = startIndex + i * taskSize;
+        size_t lastParticle        = std::min(startIndex + (i + 1) * taskSize, endIndex);
+        size_t numParticlesCompute = lastParticle - firstParticle;
 
         unsigned numThreads = 256;
```
rmcabezon, line 98: Are the numThreads tailored for Piz Daint? Are these the same amount among all cuda kernels or does each one has its own characteristic value (I saw 128 in other kernels)? Shouldn't this be something that the user can define?

rmcabezon (approved):

sph/include/sph/hydro_std/density_gpu.cu
sekelle: This is a low-level implementation detail that the user should not touch. It's not specific to Piz Daint.

PR #267, opened 2022-05-23
Update CMakeLists.txt

jfavre:
both sphexa and sphexa-cuda can now run with insitu viz enabled


jgphpc (approved):

sekelle:
ha, I had already done the same change in a local version of mine

PR #269, opened 2022-05-26
Wind shock test

jaescartin1:
Preliminary scenario with 1 cube with the bubble in the middle. The advanced scenario will contain 4 cubes follow the x-axe (the existing one + 3 more with density 1) to show the movement of the bubble by the wind.


main/src/io/ifile_writer.hpp
```diff
@@ -30,7 +30,7 @@ struct AsciiWriter : public IFileWriter<Dataset>
               std::string path) const override
     {
         const char separator = ' ';
-        path += std::to_string(d.iteration) + ".txt";
+        path += std::to_string(d.ttot) + ".txt";
```
sekelle, line 33: I think it's confusing to use floating point numbers in file names. An important use-case for these files
is to do quick checks and debugging on certain steps. If you insist on using plain text files for scientific experiments with time series, then you can find the iteration time in `constants.txt`

sekelle:
The rest of the code looks very well I think. My advice for implementing the advanced scenario as simply as possible would be to construct the extra 3 cubes separately. I.e. you keep the construction and distribution of the cube with the bubble as is and with `N` ranks, each rank additionally gets the coordinates of 1/Nth of the 3 extra cubes. This would be much easier than figuring out directly how to get 1/Nth of the composite system.

main/src/io/ifile_writer.hpp
rmcabezon: I don't think it is confusing. It is in fact something that is commonly used mostly when you want outputs at specific times and you don't want to keep looking in the table which is which. I'd appreciate having both options if possible.

rmcabezon (approved):

PR #270, opened 2022-05-27
Make VE the default propagator

sekelle:
This will make modern SPH the default choice. The previous default was standard SPH and can be activated with the
option `--prop std`

@jfavre Note that this affects the insitu visualization. In the implementation of modern SPH, the field "rho" is not used.
Instead, rho is calculated from three other fields:
```
rho_i = kx_i * m_i / xm_i;
```
If the density is required for in-situ visualization, then it will have to be produced with a filter that performs the calculation above
(or use standard SPH)


jfavre (approved):
is the choice of words "modern", vs. "standard" a decision of the SPH-EXA project, or is this widely accepted in the SPH community?

rmcabezon:
> is the choice of words "modern", vs. "standard" a decision of the SPH-EXA project, or is this widely accepted in the SPH community?

It is accepted in the SPH community. Maybe there are a few variations, but this is well understood by people that are up to date.

rmcabezon (approved):

PR #272, opened 2022-06-02
Conserved quantities

sekelle:
Conserved quantities like energies and momenta are observables, so I moved them there.


rmcabezon:
> Conserved quantities like energies and momenta are observables, so I moved them there.

Agreed!

rmcabezon (approved):

PR #274, opened 2022-06-09
Temperature equation

rmcabezon:
Added a new ideal gas EOS based on temperature instead of internal energy. There is a new corresponding initialization --init SedovTemp and a new propagator --prop temp. This is done to avoid conflicts with the develop branch and keep the old implementation for now.


sekelle:
I'm assuming that the motivation for this change is to prepare for the Helmholtz EOS?
In its current form with only the temperature ideal gas equation, it is very hard to see what the actual goal of this PR might be.

There are 3 additional conserved fields and one additional dependent field, all of which can be optimized away for the ideal gas.
* `mui` and `cv` are constants
* out of `temp` and `u`, only one is needed as U = cv * T
* `prho` and `TdpdTrho` are identical

Which of these properties are no longer true for the Helmholtz EOS?

rmcabezon:
> I'm assuming that the motivation for this change is to prepare for the Helmholtz EOS? In its current form with only the temperature ideal gas equation, it is very hard to see what the actual goal of this PR might be.

The temperature equation is more stable in the case of degenerate matter, such as in white dwarfs and neutron stars. Is going to be difficult to simulate those without it. Also, the Helmholz EOS, which should be basically our default EOS for most physical simulations takes Temperature as an input variable. As Helmholtz at low T and density behaves as ideal gas, we can have it as default. Unless it is very clear that the ideal gas is enough (f.e. for tests) we should use Helmholtz.

> There are 3 additional conserved fields and one additional dependent field, all of which can be optimized away for the ideal gas.
> * `mui` and `cv` are constants

For the tests may be, but this is not true for most simulations that take into account composition. Also, the relation between them is not the same for different EOS.

> * out of `temp` and `u`, only one is needed as U = cv * T

Yes. Which one is determined by the EOS input. As I mentioned, some EOS require T, others require U, and others require entropy, which we have not yet included.

> * `prho` and `TdpdTrho` are identical

True. Only for ideal gas.

> Which of these properties are no longer true for the Helmholtz EOS?

Already answered above.

jgphpc:
cscs-ci run

sekelle:
Addressed by https://github.com/unibas-dmi-hpc/SPH-EXA/pull/318

PR #276, opened 2022-06-15
Add readme about Ascent in a container

jgphpc:


jfavre (changes requested):
built with Ascent also requires the following module load:

cudatoolkit/11.2.0_3.39-2.1__gf93aa1c

jfavre (approved):

sekelle (approved):

PR #278, opened 2022-06-17
Fixed boundaries

lks1248:


domain/include/cstone/sfc/box.hpp
```diff
@@ -101,23 +104,33 @@ HOST_DEVICE_FUN constexpr int pbcDistance(int x)
 template<class T>
 class Box
 {
+    HOST_DEVICE_FUN void validateBoundaries(int boundaryX, int boundaryY, int boundaryZ)
+    {
+        if (boundaryX < 0 || boundaryX > 2 || boundaryY < 0 || boundaryY > 2 || boundaryZ < 0 || boundaryZ > 2)
+        {
+            throw std::runtime_error("Invalid boundary types! (not 0, 1, 2)");
+        }
+    }
```
sekelle: This should a) not be a public member function and b) not be here at all, because if you use a proper enumeration with named constants it won't be possible to specify an invalid choice.

domain/include/cstone/sfc/box.hpp
```diff
@@ -136,41 +149,37 @@ class Box
...
+    HOST_DEVICE_FUN constexpr int boundaryY() const { return boundaries[1]; } // NOLINT
+    HOST_DEVICE_FUN constexpr int boundaryZ() const { return boundaries[2]; } // NOLINT
+
+    HOST_DEVICE_FUN constexpr bool pbcX() const { return boundaries[0] == 1; } // NOLINT
+    HOST_DEVICE_FUN constexpr bool pbcY() const { return boundaries[1] == 1; } // NOLINT
+    HOST_DEVICE_FUN constexpr bool pbcZ() const { return boundaries[2] == 1; } // NOLINT
+
+    HOST_DEVICE_FUN constexpr bool fbcX() const { return boundaries[0] == 2; } // NOLINT
+    HOST_DEVICE_FUN constexpr bool fbcY() const { return boundaries[1] == 2; } // NOLINT
+    HOST_DEVICE_FUN constexpr bool fbcZ() const { return boundaries[2] == 2; } // NOLINT
```
sekelle: a) Integer numbers should never be used directly to encode different cases or options. This is what enumerations are for (`enum class`).

sph/include/sph/positions.hpp
```diff
@@ -71,7 +113,11 @@ void computePositions(size_t startIndex, size_t endIndex, Dataset& d, const csto
         T deltaA = dt + T(0.5) * dt_m1;
         T deltaB = T(0.5) * (dt + dt_m1);
 
-        Vec3T Val = (X - X_m1) * (T(1) / dt_m1);
+        Vec3T Val = (X - X_m1) * (T(1) * hasFBC / dt_m1);
+        if (Val[0] == 0 && Val[1] == 0 && Val[2] == 0)
+        {
+            // printf("val zero: i %lu hasFBC %d", i, d.hasFBC[i]);
+        }
```
sekelle: left-over junk code

sph/include/sph/positions.hpp
```diff
@@ -38,6 +38,41 @@
...
+    if (box.fbcZ())
+    {
+        T distZmax = std::abs(box.zmax() - z);
+        T distZmin = std::abs(box.zmin() - z);
+
+        if (distZmax < 2.0 * h || distZmin < 2.0 * h)
+        {
+            if (vx == T(0.0) && vy == vx && vz == vy) { return T(0.0); }
+        }
+    }
```
sekelle: code bloat: the three if statements can all be expressed with just one function and different arguments

sph/include/sph/positions.hpp
```diff
@@ -38,6 +38,41 @@
 
 namespace sph
 {
+template<class T>
+T fbcCheck(T x, T y, T z, T vx, T vy, T vz, T h, const cstone::Box<T> box)
```
sekelle: take box by const-reference
doc-string missing

sekelle: This function should just return a bool whether the particle is inside the fixed zone.
If that's the case, the rest of the integrator is simply skipped for this particle (the `continue` keyword)

sekelle (changes requested):
As an additional remark, please either revert any formatting changes in files that have no actual code changes (like gather.cu or exchange_halos.hpp) or alternatively open a separate PR for that.

domain/include/cstone/sfc/box.hpp
```diff
@@ -136,41 +154,38 @@ class Box
...
+    HOST_DEVICE_FUN constexpr BoundaryType boundaryY() const { return boundaries[1]; } // NOLINT
+    HOST_DEVICE_FUN constexpr BoundaryType boundaryZ() const { return boundaries[2]; } // NOLINT
+
+    HOST_DEVICE_FUN constexpr bool pbcX() const { return boundaries[0] == BoundaryType::periodic; } // NOLINT
+    HOST_DEVICE_FUN constexpr bool pbcY() const { return boundaries[1] == BoundaryType::periodic; } // NOLINT
+    HOST_DEVICE_FUN constexpr bool pbcZ() const { return boundaries[2] == BoundaryType::periodic; } // NOLINT
+
+    HOST_DEVICE_FUN constexpr bool fbcX() const { return boundaries[0] == BoundaryType::fixed; } // NOLINT
+    HOST_DEVICE_FUN constexpr bool fbcY() const { return boundaries[1] == BoundaryType::fixed; } // NOLINT
+    HOST_DEVICE_FUN constexpr bool fbcZ() const { return boundaries[2] == BoundaryType::fixed; } // NOLINT
```
sekelle: The `pbc` and `fbc` functions are not required anymore as you can just do:
```c++
if (box.boundaryX() == BoundaryType::periodic)
```

main/src/init/fixed_boundaries.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+#pragma once
+
+#include "cstone/sfc/box.hpp"
+#include "sph/particles_data.hpp"
+
+namespace sphexa
+{
+template<class T, class Dataset>
+void applyFixedBoundaries(Dataset& d, cstone::Box<T>& box)
```
sekelle: * This should not take the whole Dataset, just one out of (x,y,z) and all of (vx, vy, vz)
* similar simplification as in updatePosition applies

sekelle: Actually, as you are using initialization by file, this should not be done in C++.
It's not a clean style to parse some of the IC from file, but then also do some further init in C++ on top.
You should very easily be able to correctly initialize the velocities in the external program that generated the file,
then you don't have to mix two different styles of initialization.

sekelle:
cscs-ci run

sekelle (approved):

PR #282, opened 2022-07-21
Prevent neighbor list access out of bounds

sekelle:


jgphpc (approved):
ci offline, ran tests manually

PR #283, opened 2022-07-26
GPU-direct halo exchange

sekelle:
* Added user-friendly interface to refer to particle fields based on compile-time strings
* Support for GPU-direct halo exchange

GPU-direct is off by default and needs to be turned by passing the option `-DGPU_DIRECT=ON` to cmake.
On Cray, at least `cray-mpich/7.7.20` contained in `cdt/22.05` is required as prior versions of cray-mpich are too buggy.


jgphpc (approved):
tested manually, all unit tests passed with 4 container images,
where CMAKE_BUILD_TYPE=Release|Debug and -DGPU_DIRECT=ON|OFF
Images pushed to art.cscs.ch/contbuild/testing/jg/sph-exa_install

analytical tests result will be added later but this can be merged.

PR #287, opened 2022-08-12
Hip readiness

sekelle:
This is purely a clean-up PR to remove some code duplication and ensure that only `*.cuh` and `*.cu` have to be hipified.


jgphpc (approved):
tested without gitlab on daint with jfrog.svc.cscs.ch/contbuild/testing/jg/sph-exa_build:pr287
all tests passed except integration_mpi/exchange_halos_gpu (need rdma config inside container)
-> ready to be merged.

PR #289, opened 2022-08-13
Restart capability for turbulence runs

sekelle:
This takes care of initializing the turbulence modes and amplitudes when initializing the hydro data from file,
including RNG state and phases.

Additional changes
* default HDF5 output file suffix changed from `.h5part` to `.h5`
* added `-o` switch to specify the output file. The default is unchanged `dump_...`.
* The default output field list (`-f`) is now the list of conserved fields, such that output files are by default restartable.

Example: Turbulence simulation up to 18s with 2 restarts:
```
./sphexa-cuda --glass glass.h5 --prop turbulence -n 150 --init turbulence -s 6.0 -w 6.0 -o turb.6s.h5
./sphexa-cuda --glass glass.h5 --prop turbulence --init turb.6s.h5 -s 12.0 -w 12.0 -o turb.12s.h5
./sphexa-cuda --glass glass.h5 --prop turbulence --init turb.12s.h5 -s 18.0 -w 18.0 -o turb.18s.h5
```
The final snapshot at t = 18s will be in the file `turb.18s.h5`.
To compute additional dependent quantities, like density, pressure, curlv, etc, we can restart from `turb.18s.h5`, do one
additional SPH iteration and save the fields we want to a different file, e.g.:
```
./sphexa-cuda --glass glass.h5 --prop turbulence --init turb.18s.h5 -s 18.001 -w 18.001 -f x,y,z,rho,curlv -o turbulence.h5
```


rmcabezon:
> ./sphexa-cuda --glass glass.h5 --prop turbulence --init turb.18s.h5 -s 18.001 -w 18.001 -f x,y,z,rho,curlv -o turbulence.h5

The problem I see with the command above is that it doesn't ensure that just one step will be done. You will need to check with the current timestep or the initial timestep that you use.

Alternatively, could this last command be as follows?
./sphexa-cuda --glass glass.h5 --prop turbulence --init turb.18s.h5 -s 1 -w 1 -f x,y,z,rho,curlv -o turbulence.h5

Or does a number of steps relative to the last update don't work? Would we need to know the actual number of steps performed until 18 s?

main/src/init/factory.hpp
```diff
@@ -54,13 +56,23 @@ std::unique_ptr<ISimInitializer<Dataset>> initializerFactory(std::string testCas
     if (testCase == "sedov")
     {
         if (glassBlock.empty()) { return std::make_unique<SedovGrid<Dataset>>(); }
+#ifdef SPH_EXA_HAVE_H5PART
```
rmcabezon, line 59: Why is the ifdef still mentioning H5PART instead of H5 or HDF5?

rmcabezon: Ok. I see. To get rid of this you will need to do too many changes in the code. Not worthy.

main/src/sphexa/sphexa.cpp
```diff
@@ -187,8 +186,8 @@ void printHelp(char* name, int rank)
         printf("%s [OPTIONS]\n", name);
         printf("\nWhere possible options are:\n\n");
 
-        printf("\t--init \t\t Test case selection (evrard, sedov, noh, isobaric-cube, wind-shock) or an HDF5 file "
-               "with initial conditions\n");
+        printf("\t--init \t\t Test case selection (evrard, sedov, noh, isobaric-cube, wind-shock turbulence)\n"
```
rmcabezon: A comma is missing before turbulence.

sph/include/sph/hydro_turb/driver.hpp
```diff
@@ -92,16 +93,17 @@ void updateNoise(std::vector<T>& phases, T stddev, T dt, T ts, std::mt19937& gen
 
 /*! @brief Adds the stirring motion to particle accelerations
  *
- * @tparam Dataset
- * @param  startIndex
- * @param  endIndex
- * @param  d
+ * @tparam Dataset          SPH hydro dataset
+ * @tparam TurbulenceModes  Turbulence stirring modes
```
rmcabezon: Do you mean turb.modes? I don't see TurbulenceModes anywhere.

rmcabezon (approved):

jgphpc (approved):
tested (without gitlab) on daint with jfrog.svc.cscs.ch/contbuild/testing/jg/sph-exa_build:pr289
🆗 
```
ci_gputests %unittest=noh
   - builtin
      * num_tasks: 1
      * Density: 0.9556674971872787
      * Pressure: 0.3849719983304184
      * Velocity: 0.03835330256298284
------------------------------------------------------------------------------
ci_gputests %unittest=sedov
   - builtin
      * num_tasks: 1
      * Density: 0.13760054829121357
      * Pressure: 0.9024732492140309
      * Velocity: 0.9149740287440481
```

main/src/sphexa/sphexa.cpp
sekelle: > Or does a number of steps relative to the last update don't work? Would we need to know the actual number of steps performed until 18 s?

Yes, you need to know the last iteration and add 1, either from the program output or by using one of the scripts in the repository in `scripts/` for HDF5 that will print out the contained steps in the file.

sph/include/sph/hydro_turb/driver.hpp
sekelle: Left-over template parameter, I'll remove it.

PR #292, opened 2022-08-17
Ryoanji demostrator app on LUMI

sekelle:


jgphpc (approved):
Tested (without gitlab) on dom with:
jfrog.svc.cscs.ch/contbuild/testing/anfink/9590261141699040/pasc/sphexa/sph-exa_build:pr292
:ok:

PR #297, opened 2022-08-27
Wind blob test extension

lks1248:
-Added observable
-extended Domain to avoid collision through PBC


jgphpc:
ci was failing because of an older yml. merging develop solved the issue.

sekelle (approved):

PR #299, opened 2022-09-07
Update catalyst_adaptor.h

jfavre:
point connectivity list can be copied via shared memory pointer


jgphpc:
Can you add this ?
```
#include "init/wind_shock_init.hpp"
```
in main/src/sphexa/sphexa.cpp before
```
#include "init/factory.hpp"
```

sekelle:
> Can you add this ?
> 
> ```
> #include "init/wind_shock_init.hpp"
> ```
> 
> in main/src/sphexa/sphexa.cpp before
> 
> ```
> #include "init/factory.hpp"
> ```

Is this still pending?

main/src/sphexa/sphexa.cpp
```diff
@@ -43,6 +43,7 @@
 #endif
 
 #include "cstone/domain/domain.hpp"
+#include "init/wind_shock_init.hpp"
 #include "init/factory.hpp"
```
sekelle: Actually, I don't see why this include is needed as `init/factory.hpp` includes `init/wind_shock_init.hpp` if HDF5 support is enabled. If you include this here, I would expect it to break the non-HDF5 build.

sekelle (approved):

PR #310, opened 2022-10-10
Integration of nuclear networks

jolatechno:
Added the Nuclear Network submodule:
- Made sure that it would be included in the Cmake build process.
- Made sure that SPH-EXA would still compile without the nuclear network submodule.
- Started adding the nuclear data container to simulation data.
- Some boiler plate nuclear propagator and nuclear energy observable.


jolatechno:
Edit: my bad about the formating, I didn't know what the right clang formating was.

LICENSE
```diff
@@ -18,4 +18,4 @@ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
 AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-SOFTWARE.
```
sekelle: Spurios change?

CMakeLists.txt
```diff
@@ -100,4 +110,4 @@ add_subdirectory(domain)
 add_subdirectory(ryoanji)
 add_subdirectory(sph)
 add_subdirectory(physics)
-add_subdirectory(main)
```
sekelle: Spurious change?

main/src/init/nuclear_sedov_init.hpp
```diff
@@ -0,0 +1,208 @@
...
+#include "nnet/sphexa/initializers.hpp"
+
+#include "nnet/parameterization/net87/net87.hpp"
+#include "nnet/parameterization/net14/net14.hpp"
+
+namespace sphexa
+{
+
+template<class Dataset>
+class NuclearSedovGrid : public ISimInitializer<Dataset>
```
sekelle: I don't think we need to bother with a regular grid init for the SPH part when using nnet. This option is only there to run a minimal example if HDF5 is absent for some reason. This is never useful for more elaborate workload, such as when using nnet.

main/src/observables/factory.hpp
```diff
@@ -39,6 +39,9 @@
 #include "time_energies.hpp"
 #include "gravitational_waves.hpp"
 #include "wind_bubble_fraction.hpp"
+#ifdef USE_NUCLEAR_NETWORKS
+#include "nuclear_observable.hpp"
+#endif
```
sekelle, line 44: Don't you want to register the new nuclear_observable in the factor in this file?

main/src/propagator/factory.hpp
```diff
@@ -59,6 +62,24 @@ propagatorFactory(const std::string& choice, size_t ngmax, size_t ng0, std::ostr
...
+#ifdef USE_NUCLEAR_NETWORKS
+        return std::make_unique<NuclearProp<DomainType, ParticleDataType, 86, false>>(ngmax, ng0, output, rank);
+#endif
+    }
+    if (choice == "std-net87")
+    {
+#ifdef USE_NUCLEAR_NETWORKS
+        return std::make_unique<NuclearProp<DomainType, ParticleDataType, 87, false>>(ngmax, ng0, output, rank);
+#endif
+    }
```
sekelle: Coalescing the macro into a single block would probably look nicer visually.

sekelle:
As a general remark, I would not put the `NuclearData` class inside the nnet submodule. Instead, in `physics/nnet` you could have two folders, one for the submodule, and another one for the interfaces to the submodule. `NuclearData` is, effectively, one of those interfaces, as it is modeled according to the requirements of the SPH-EXA application front-end and has header dependencies on the Cornerstone library.

If you have a choice, I would not introduce this dependency on Cornerstone into your nnet code. As an example, you can look at the gravity solver Ryoanji. We don't pass ParticlesData, or any Octree class from Cornerstone to Ryoanji. Instead, we pass the coordinate fields, masses and the tree as elementary arrays of floating point or integer types. This allows as to swap it out for something else if we need to, as long as the storage format stays the same. And we enable somebody else to call Ryoanji
who might not be using ParticlesData or Cornerstone to build octrees.

Another analogy could be BLAS matrix-matrix multiplication (dgemm). Rather than a custom matrix class, the interface accepts floating point arrays in a specific format: column or row major.

LICENSE
jolatechno: Yes I will revert this change in the next commit.

CMakeLists.txt
jolatechno: some changes are necessary, other were done on accident and will be reverted. The last line change is very weird as it's not present in my local repo.

main/src/init/nuclear_sedov_init.hpp
jolatechno: Ok I will remove the grid propagator then.

main/src/observables/factory.hpp
jolatechno: There is no real nuclear test case so I don't know where to reference nuclear_observable in the factory, but it will be done in the future. I don't think IO is the first thing to do in the integration.

main/src/propagator/factory.hpp
jolatechno: Ok I will move to a single ifdef if.

main/src/init/sedov_constants.hpp
```diff
@@ -14,11 +14,12 @@ std::map<std::string, double> sedovConstants()
...
         {"r0", 0.},
         {"r1", 0.5},
-        {"mTotal", 1.},
-        {"energyTotal", 1.},
+        {"mTotal", 1e9},
+        {"energyTotal", 1e9},
+        {"nuclearTemperature", 1e9},
         {"width", 0.1},
-        {"rho0", 1.},
-        {"u0", 1e-8},
```
sekelle: This breaks the existing Sedov test case.

sekelle (changes requested):

main/src/init/sedov_constants.hpp
jolatechno: With these constants (curently used in this branch) the sedov test case works for me, which parameter causes a bug for you ?

`
        {"mTotal", 1e9},
        {"energyTotal", 1e9},
        {"nuclearTemperature", 1e9},
        {"width", 0.1},
        {"rho0", 1e9},
        {"u0", 1e1},
`

sekelle: The ones that you changed of course: `rho0` and `u0`. This affects the sedov test (`--init sedov`) for which we have an analytical solution.

rmcabezon: Be both aware that these constants are specific for an LB test. They have no generality for any other thing. Including nuclear networks themselves. Therefore, they probably shouldn't be included in any of the standard tests that we have. Maybe in an independent init file?

main/src/init/nuclear_sedov_init.hpp
sekelle: This is still pending.

sekelle:
some of the TODOs still pending

main/src/init/nuclear_sedov_init.hpp
jolatechno: When I remove the grid propagator I always get an error, do you mean that the glass propagator is not needed ? I don't properly understand the difference between the grid and glass propagators.

jolatechno:
> some of the TODOs still pending

So I fiexed the issue you had with `dt` being initalized in `nuclear_data.hpp`, and as I commented earlier, `sedov-nuclear` stops working if I remove the `NuclearSedovGrid` initializer.

Apart from the first point which is resolved, and the second for which I will need some clarifications, what TODOs do you still see ?

jolatechno:
I didn't properly understand the difference between `grid` and `glass` initilaizers, now the `NuclearSedovGrid` has been removed and I think all TODOs are resolved.

sekelle:
I've updated this branch to include the latest version in the `develop` branch and updated the submodule version of the SPH-EXA-nnet repository.

The main problem is still that all of the nuclear-network code is included via headers into the compilation unit of the SPH-EXA main function. At the level of the propagator, the call to ` sphnnet::computeNuclearReactions` should be made through a library interface.

jolatechno:
Hello, obviously I understand that this merge request was closed as the branch is out of date. Are there any plans to still integrate nuclear networks to the main branch ?

PR #318, opened 2022-10-21
Temperature equation

sekelle:
* `mui` is now listed as a constant for those test cases that only use a single value for all particles
* the integrator has been generalized to deal with future test cases that use different values of `mui` per particle
* total internal energy is still computed, but is not active as a field anymore. Where needed, the internal energy is computed as `u = cv * T`


main/src/init/noh_init.hpp
```diff
@@ -55,23 +62,27 @@ void initNohFields(Dataset& d, double totalVolume, const std::map<std::string, d
...
+    d.minDt_m1  = firstTimeStep;
+
+    auto cv    = sph::idealGasCv(d.muiShared);
+    auto temp0 = constants.at("u0") / cv;
+
     std::fill(d.m.begin(), d.m.end(), mPart);
     std::fill(d.h.begin(), d.h.end(), hInit);
     std::fill(d.du_m1.begin(), d.du_m1.end(), 0.0);
-    std::fill(d.mui.begin(), d.mui.end(), 10.0);
+    std::fill(d.mui.begin(), d.mui.end(), constants.at("mui"));
```
rmcabezon: Why the difference in initializing mui to a constant value in the Noh test, while using muiShared in the other tests?

rmcabezon (commented):
muiShared is a bit weird name. Would it be possible to change it to a more physical meaning, such as mui_constant ?

jolatechno:
To be clear, for the Helmholtz EOS `u != cv * T` (because `cv` varies with `T`), so `u(T,  rho, Y)` is directly computed by the Helmholtz EOS and `cv(T,  rho, Y) = du/dT` (also computed by the helmholtz EOS).

Will that cause a problem for the curent "tempearture equation" ?

main/src/init/noh_init.hpp
sekelle: The value is the same, as `muiShared = constants.at("mui")`.
I'll change this for consistency and rename `muiShared` to `muiConst`

sekelle:
> Will that cause a problem for the current "temperature equation" ?

No, this PR -- while not implementing the Helmholtz EOS -- will minimize the amount of further changes needed.

rmcabezon (approved):

PR #324, opened 2022-11-07
Moved TuplePlus to util namespace for reuse

sekelle:
`TuplePlus` can be included with `cstone/util/tuple.hpp`. It works for arbitrary `thrust` and `std` tuples in host and device code.


lks1248 (approved):

jgphpc:
retry after merging https://github.com/unibas-dmi-hpc/SPH-EXA/pull/326

PR #332, opened 2022-11-22
Added functionality to read internal energy or temperature from file

lks1248:


main/src/init/file_init.hpp
```diff
@@ -135,9 +162,9 @@ cstone::Box<typename HydroData::RealType> restoreHydroData(const std::string& h5
     errors |= fileutils::readH5PartField(h5_file, "z", d.z.data());
     errors |= fileutils::readH5PartField(h5_file, "h", d.h.data());
     errors |= fileutils::readH5PartField(h5_file, "m", d.m.data());
-    errors |= fileutils::readH5PartField(h5_file, "temp", d.temp.data());
+    errors |= initUTemp(h5_file, rank,d);
```
sekelle, line 165: It would be better to convert any left-over checkpoint HDF5 files from `u` to `temp` with a python script and simply require `temp` as input. The version here is not that useful as it won't be able to adapt to different equations of state. For a more general EOS with u != cv * temp, you will have both `temp` and `u`. Here you'll try to load both and throw an error one level above in the test-case specific initializer if the file didn't contain the required information.

main/src/io/mpi_file_utils.hpp
```diff
@@ -201,6 +201,20 @@ void writeH5Part(Dataset& d, size_t firstIndex, size_t lastIndex, const cstone::
...
+    {
+        d.release("c22");
+        d.acquire("u");
+        T cv = sph::idealGasCv(d.muiConst);
+#pragma omp parallel for schedule(static)
+        for(size_t i = firstIndex; i < lastIndex; ++i)
+        {
+            d.u[i] = d.temp[i] * cv;
+        }
+    }
```
sekelle, line 217: Other than the missing release `u` and re-acquirement of `c22` , this code is fine. But it belongs into ve_hydro.hpp `prepareOutput` and `finishOutput` where all the other output-related data tweaks are and where you have access to the selected EOS.

sekelle (changes requested):

PR #340, opened 2022-12-06
gpu observables 

lks1248:


lks1248:
Added GPU-only calculations for the Kelvin-Helmholtz-, Turbulence- and Wind-Shock tests.

sekelle (approved):

PR #341, opened 2022-12-07
Mach RMS calculations for turbulence runs

lks1248:


main/src/observables/conserved_gpu.cu
```diff
@@ -59,42 +59,43 @@ struct EMom
...
-    return {0.5 * get<0>(result), get<1>(result), get<2>(result), get<3>(result)};
+    return {0.5 * get<0>(result), get<1>(result), get<2>(result), get<3>(result), get<4>(result)};
 }
 
 #define CONSERVED_Q_GPU(Tc, Tv, Tt, Tm)                                                                                \
-    template std::tuple<double, double, Vec3<double>, Vec3<double>> conservedQuantitiesGpu(                            \
+    template std::tuple<double, double, Vec3<double>, Vec3<double>, double> conservedQuantitiesGpu(                    \
         Tt, const Tc* x, const Tc* y, const Tc* z, const Tv* vx, const Tv* vy, const Tv* vz, const Tt* temp,           \
-        const Tm* m, size_t, size_t)
+        const Tm* m, const Tc* c, size_t, size_t)
```
sekelle: This should be disentangled here to have some flexibility and options to compose different observable.
The mach numbers should be handled with a separate function here.
What you can reuse/extend for now are the `conserved_gpu` files and the `computeConservedQuantities` where the MPI reduction happens.

main/src/observables/conserved_quantities.hpp
```diff
@@ -79,9 +81,10 @@ auto localConservedQuantities(size_t startIndex, size_t endIndex, Dataset& d)
         eInt += cv * temp[i] * mi;
         linmom += mi * V;
         angmom += mi * cross(X, V);
+        machSum += norm2(V) / (c[i]*c[i]);
```
sekelle: Also the CPU should do this computation in a separate function.

sekelle (changes requested):

main/src/observables/conserved_gpu.cu
```diff
@@ -91,6 +91,32 @@ conservedQuantitiesGpu(Tt cv, const Tc* x, const Tc* y, const Tc* z, const Tv* v
     return {0.5 * get<0>(result), get<1>(result), get<2>(result), get<3>(result)};
 }
 
+template<class Tc, class Tv>
+struct MachSum
```
rmcabezon: I would call this Mach2Sum or MachSqSum, as we are in fact calculating the squared value of the Mach number.

main/src/observables/conserved_gpu.cu
```diff
@@ -91,6 +91,32 @@ conservedQuantitiesGpu(Tt cv, const Tc* x, const Tc* y, const Tc* z, const Tv* v
...
+};
+
+template<class Tc, class Tv>
+double machSumGpu(const Tv* vx, const Tv* vy, const Tv* vz, const Tc* c, size_t first, size_t last)
+{
+    auto it1 = thrust::make_zip_iterator(thrust::make_tuple(vx + first, vy + first, vz + first, c + first));
+    auto it2 = thrust::make_zip_iterator(thrust::make_tuple(vx + last, vy + last, vz + last, c + last));
+
+    auto plus         = thrust::plus<double>{};
+    auto localMachSum = 0.0;
```
rmcabezon: Same as above. localMach2Sum or localMachSqSum

main/src/observables/conserved_gpu.cu
```diff
@@ -91,6 +91,32 @@ conservedQuantitiesGpu(Tt cv, const Tc* x, const Tc* y, const Tc* z, const Tv* v
...
+    double operator()(const thrust::tuple<Tv, Tv, Tv, Tc>& p)
+    {
+        Vec3<double> V{get<0>(p), get<1>(p), get<2>(p)};
+        Tc           c = get<3>(p);
+        return norm2(V) / (double(c) * c);
+    }
+};
+
+template<class Tc, class Tv>
+double machSumGpu(const Tv* vx, const Tv* vy, const Tv* vz, const Tc* c, size_t first, size_t last)
```
rmcabezon: Same as above. Mach2SumGpu or MachSqSumGpu.

rmcabezon (commented):
The calculation is correct, but I would use Mach2 or MachSq whenever necessary (localsum, local values, etc...) to be clear that what we are calculating is in fact the squared value of the Mach number. I only pointed this on the first file, but if corrected, needs to be done in the other files, too. 
Not for MachRMS. That's correct.

sekelle (approved):

PR #349, opened 2022-12-15
Av cleaner

rmcabezon:
Includes the AV cleaner and the usage of real data for SPH kernels unit tests.


sph/include/sph/hydro_ve/momentum_energy_gpu.cu
```diff
@@ -127,7 +131,6 @@ void computeMomentumEnergy(size_t startIndex, size_t endIndex, unsigned ngmax, D
 }
 
 template void computeMomentumEnergy(size_t, size_t, unsigned,
```
jgphpc: ci log shows a build failure [here](https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/44692846847247/4264416954089684/-/jobs/3480066509#L183), if you can't see it, here is the err msg:
```
hydro_ve/momentum_energy_gpu.cu(133): error: 
no instance of function template "sph::cuda::computeMomentumEnergy" 
matches the specified type
```

I can reproduce the failure on daint, see /scratch/snx3000/piccinal/pr349/0
Maybe line 130 was wrongly deleted ?

jgphpc (changes requested):

sph/include/sph/hydro_ve/momentum_energy_gpu.cu
sekelle: It was indeed, I'll fix it.

jgphpc (approved):

sph/include/sph/hydro_ve/momentum_energy_kern.hpp
```diff
@@ -69,11 +68,9 @@ HOST_DEVICE_FUN T avRelVelCorrection(util::array<Tc, 3> R, Tc dist, Th hi, Th hj
     if (eta_ab < eta_crit) { dmy3 = std::exp(-math::pow((eta_ab - eta_crit) / T(0.2), 2)); }
     T phi_ab = T(0.5) * stl::max(T(0), stl::min(T(1), T(4) * A_ab / math::pow(T(1) + A_ab, 2))) * dmy3;
 
-    auto vAVi = Vi - phi_ab * gradV_i_dot_R;
-    auto vAVj = Vj + phi_ab * gradV_j_dot_R;
-    T    rv   = dot(R, vAVi - vAVj);
-
-    return rv;
+    // additive AV correction to rv = dot(R, Vij)
+    T rv_AV = dot(R, -phi_ab * (gradV_i_dot_R + gradV_j_dot_R));
```
rmcabezon: Shouldn't be a minus here?: 
gradV_i_dot_R - gradV_j_dot_R 

instead of:
gradV_i_dot_R + gradV_j_dot_R

sph/include/sph/hydro_ve/momentum_energy_kern.hpp
```diff
@@ -62,7 +62,7 @@ HOST_DEVICE_FUN T avRelVelCorrection(util::array<Tc, 3> R, Tc dist, Th hi, Th hj
     T A_ab = T(0);
 
     if (dmy2 != T(0)) { A_ab = dmy1 / dmy2; }
-    T eta_ab = stl::min(dist / hi, dist / hj);
+    T eta_ab = dist * stl::min(hiInv, hjInv);
```
rmcabezon: You can pass directly dist/hi and dist/hj and avoid passing one of the three. Just pass v1 and v2, which are the typical name for the scaled distances between particles, and are already calculated when calling the kernel a few lines before.

sph/include/sph/hydro_ve/momentum_energy_kern.hpp
sekelle: No, both terms where originally subtracted from V_i and V_j. We can be sure by trying it out and verify that this would make the tests fail.

sekelle (approved):

PR #351, opened 2022-12-22
Initial Ewald CPU implementation and tests.

jpcoles:
A CPU implementation of periodic boundaries using the Ewald summation method. The main entry point is the function computeGravityEwald(), which takes the same arguments are computeGravity().


jpcoles:
Correcting failed test and updating test comments.

PR #352, opened 2022-12-23
Initial Ewald CPU implementation and tests

jpcoles:
A CPU implementation of periodic boundaries using the Ewald summation method. The main entry point is the function computeGravityEwald(), which takes the same arguments are computeGravity().

### DIFF
--- a/ryoanji/src/ryoanji/nbody/traversal_ewald_cpu.hpp
+++ b/ryoanji/src/ryoanji/nbody/traversal_ewald_cpu.hpp
@@ -1,0 +1,394 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2021 CSCS, ETH Zurich
+ *               2021 University of Basel
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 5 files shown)


ryoanji/src/ryoanji/nbody/traversal_ewald_cpu.hpp
```diff
@@ -0,0 +1,600 @@
...
+template<class T1, class T2>
+void computeGravityGroupEwald(
+        const CartesianQuadrupole<T2>& Mroot, const cstone::SourceCenterType<T1>& Mroot_center, double L, 
+        std::vector<struct EwaldData> &ed_list,
+        double ewaldCut, int numShells, double alpha_scale,
+        const T1* x, const T1* y, const T1* z,
+        T1* ax, T1* ay, T1* az, T1* ugrav,
+        LocalIndex numParticles,
+        int gravity_switch
+        )
```
sekelle: We should have this function minus the `j`-loop and implement the loop one level above this function. We'll pass distance from root-center instead of coordinate arrays and return acceleration contribution by value. Then we'll be able to call this function computing the long-range correction also from a GPU kernel.

ryoanji/src/ryoanji/nbody/traversal_ewald_cpu.hpp
```diff
@@ -0,0 +1,600 @@
...
+ *
+ * Note: acceleration output is added to destination
+ */
+template<class MType, class T1, class T2, class Tm>
+void computeGravityGroupReplica(TreeNodeIndex groupIdx, const TreeNodeIndex* childOffsets, const TreeNodeIndex* internalToLeaf,
+                                const cstone::SourceCenterType<T1>* centers, MType* multipoles, const LocalIndex* layout,
+                                const T1 dx_target, const T1 dy_target, const T1 dz_target,
+                                const T1* x, const T1* y, const T1* z, const T2* h, const Tm* m, float G, T1* ax, T1* ay,
+                                T1* az, T1* ugrav)
+{
```
sekelle: If we change the signature here to directly take `targetCenter` and `targetSize`, also for the existing variant of this function, and compute targetCenter and targetSize at the call site, we only need one version of this function to traverse both the center box and replicas.

sekelle (changes requested):

ryoanji/src/ryoanji/nbody/traversal_ewald_cpu.hpp
jpcoles: ok, I'll do this.

ryoanji/src/ryoanji/nbody/traversal_ewald_cpu.hpp
jpcoles: computeGravityGroupReplica() differs from computeGravityGroup() in more ways than just the value of targetCenter. First, the P2P loop assumes that all particles in the target leaf interact with the particles in the replica leaf. This is different than the normal case where the target must not be its own source. Second, I need to apply {dx,dy,dz}_target to each target particle (in both P2P and M2P). One could pass a new array of target x,y,z values, but then a new function would still be required and would, I feel, overcomplicate the calling code.

jpcoles:
I will do a squash rebase into a single commit on all commits related to ewald once it is approved. Please don't do the rebase yet.

sekelle (dismissed):
From my side, this is ready now for squash & merge.

sekelle:
@jpcoles will need to recreate this commit himself, please nobody merge this yet.

PR #353, opened 2023-01-09
Field-by-field callable I/O interface

sekelle:
Updated to I/O interface to get the following benefits
* support for writing individual fields whenever they are available, instead of needed all output fields simultaneously
* symmetric read and write for attributes, config parameters, etc: the code that writes them is also the code that reads them = no chance of forgetting to update one without the other
* support for including GRACKLE and nuclear fields in the output (in a follow up PR)
* Attribute types in startup-files are now type-checked upon initialization

Changes in the file structure affecting startup-files:
* the "step" attribute to store the SPH iteration of a snapshot is now called "iteration" for consistency with the C++ code and to avoid confusion with the HDF5 step index
* the type of the three BoundaryType attributes changed from int32 to int8
* To start a simulation, the required fields are what is listed as conserved in the chosen propagator. This change is needed to keep a clean and safe design while extending the init support to cooling and nuclear reactions. In practice, this means that `x_m1,y_m1,z_m1,du_m1` and `alpha` now need to be present.

Under scripts, there's now `init_file.py`, a simple python script that generates a valid startup file that you may use to quickly produce startup files from arbitrary data sources.


lks1248 (approved):

rmcabezon (commented):
Do we actually need to save x_m1, y_m1, z_m1 and u_m1 in the HDF5 files? These are increasing a lot the size of the file and we don't actually need them. On the first iteration of the simulation (either real start or restart) we can estimate them in the update module given that we have an initial timestep and the rates of change of each one of those magnitudes. After the first iteration then we can keep the real values, but there is no need to save them in the file.

sekelle:
The `_m1` fields actually don't take that much space because they are stored in 32-bit. One particle takes 92-bytes with `_m1` and 76-bytes without, i.e. about a 20% difference. If you remove them, the downside is that the information loss causes the restarted simulation to differ from the one you saved. From a software correctness point of view and for giving the guarantee that store/load through a checkpoint file does not introduce any errors, this would complicate matters.

But while I think the `_m1` files should be present in the short lived checkpoint files, there's nothing stopping us from dropping these fields in the final output file that you might want to store for a longer time.

PR #359, opened 2023-02-10
Coupled smoothing-length / neighbor-count update.

sekelle:
Also changed exponents in the h-update to:

```c++
template<class T>
HOST_DEVICE_FUN T updateH(unsigned ng0, unsigned nc, T h)
{
    constexpr T c0  = 1023.0;
    constexpr T exp = 1.0 / 10.0;
    return h * T(0.5) * std::pow(T(1) + c0 * ng0 / T(nc), exp);
}
```

### DIFF
--- a/sph/include/sph/find_neighbors.hpp
+++ b/sph/include/sph/find_neighbors.hpp
@@ -16,19 +16,28 @@ void findNeighborsSph(const T* x, const T* y, const T* z, T* h, LocalIndex first
 
     unsigned ngmin = ng0 / 4;
 
-#pragma omp parallel for
+    size_t        numFails     = 0;
+    constexpr int maxIteration = 10;
+
+#pragma omp parallel for reduction(+ : numFails)
     for (LocalIndex i = 0; i < numWork; ++i)
     {
         LocalIndex id = i + firstId;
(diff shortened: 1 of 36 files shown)


sph/include/sph/find_neighbors.hpp
```diff
@@ -5,13 +5,43 @@
...
+        nc[i]         = findNeighbors(id, x, y, z, h, treeView, box, ngmax, neighbors + i * ngmax);
+
+        int iteration    = 0;
+        int maxIteration = 10;
+        while (ngmin > nc[i] || nc[i] > ngmax && iteration++ < maxIteration)
+        {
+            h[id] = updateH(ng0, nc[i], h[id]);
+            nc[i] = findNeighbors(id, x, y, z, h, treeView, box, ngmax, neighbors + i * ngmax);
+        }
+    }
```
sekelle, line 41: Here's the main algorithmic change on the CPU. On the GPU it's implemented the same.

rmcabezon: Could we get a warning when particles do not converge? (i.e. when we reach iteration = 10 and still needs more)

rmcabezon (approved):

sph/include/sph/find_neighbors.hpp
sekelle: done

PR #368, opened 2023-03-15
Update README.md

rmcabezon:
Updated the README with missing information


rmcabezon:
I know that more has to be said about the test cases and the output fields, but at least this gives minimum usable information.

sekelle (approved):

PR #374, opened 2023-03-22
Profiling

OsmanSeckinSimsek:
This adds the profiling support to the timer for getting per-function timings in each propagator. Additionally, the halo synchronisation now returns timestamps collected after each mpi call which is then combined with the profiler timings.


domain/include/cstone/halos/exchange_halos_gpu.cuh
```diff
@@ -109,6 +117,10 @@ void haloExchangeGpu(int epoch,
         mpiSendGpuDirect(sendPtr, sendBytes, destinationRank, haloExchangeTag, sendRequests, sendBuffers);
         sendPtr += sendBytes;
     }
+    tlast      = Clock::now();
+    timings[0] = std::chrono::duration_cast<Time>(tlast - tstart).count();
+    // std::cout << "mpiSendAsync = " << timings[0] << std::endl;
+    tstart = tlast;
```
sekelle: This is an asynchronous function call, the timings that you record here will be wrong; this is not measuring the time it took to send.

domain/include/cstone/halos/exchange_halos_gpu.cuh
```diff
@@ -171,6 +186,7 @@ void haloExchangeGpu(int epoch,
     // MUST call MPI_Barrier or any other collective MPI operation that enforces synchronization
     // across all ranks before calling this function again.
     // MPI_Barrier(MPI_COMM_WORLD);
+    return std::tuple<float, float>{timings[0], timings[1]};
```
sekelle: Out of these two values, the first one is wrong and the second one corresponds to the time spent in the entire function which you can observe at the call site. This means you're not getting any added value for this manual instrumentation.

sekelle (changes requested):
What I would suggest to do is simply to replace the current MasterProcessTimer with the new Profiling timer,  recording the time-stamps on all ranks. Then you won't have to change the propagators at all, except adding a method in the BasePropagator to gather the results and write to file which you could call from main at the end of the simulation if the --profile flag was set.

domain/include/cstone/halos/exchange_halos_gpu.cuh
OsmanSeckinSimsek: I am not actually trying to measure the time it takes for the async messages to be sent. There are two timings measured here essentially, the first one is the time it takes before the mpiRecvSync call (this timing is not very important since the process is not waiting idly) and the second timing that measures the while loop in which we wait all the messages to arrive (in here the process is not doing something significant, but waiting for synchronisation. This is exactly the part I'm trying to measure). After the mpiRecvSync, we have an mpi_waitall call and the timing for that will be captured when the propagator call is made.

The reason I am returning both measurements is that it is easier to plot them later on on the horizontal bars. In the first timing until the sendAsync calls finish, the process is doing something, in the second timing where we wait for messages, and the waitAll function call is captured once we call the profilingTimer from the propagator. Waitall is the one that actually waits for the messages to be sent and this is overlapped with the mpiRecvSync, so we are not waiting twice.

It is fair to not include these manual instrumentations in the production code. Do you think it is better to keep the propagator profiling calls in the production and have the manual instrumentations in a separate branch?

sekelle: True, the second value you're returning could be different from the total as it might be necessary to wait for the sends to be completed. But the first value measures the time it takes for buffer setup and the asynchronous MPI_Isend calls to return, which is not well defined as it's not clear how much overlap there is between communication and preparing the buffers between the first to time-stamps.

It doesn't hurt to have the propagator profiling calls in the main production branch, as long as they replace the existing `timer.step` calls that only measure `rank 0` (instead of complementing them).

sekelle: > is fair to not include these manual instrumentations in the production code. Do you think it is better to keep the propagator profiling calls in the production and have the manual instrumentations in a separate branch?

Yes, I think so.

domain/include/cstone/halos/exchange_halos_gpu.cuh
```diff
@@ -33,6 +33,7 @@
 
 #include <numeric>
 #include <vector>
+#include <chrono>
```
sekelle: no longer needed

main/src/util/profiler.hpp
```diff
@@ -0,0 +1,147 @@
...
+                    funcDurations[r] = durations[r * funcTimes.size() + i];
+                }
+                timeSteps.push_back(funcDurations);
+            }
+        }
+    }
+
+    // LiB metrics calculations
+    void calculateMetrics(float* durations)
+    {
```
sekelle, line 148: This analysis part could be done in Python. I believe it would be more flexible and make it easier to adapt to changing needs in the future.

main/src/util/profiler.hpp
```diff
@@ -0,0 +1,147 @@
...
+        funcTimes.push_back(duration);                 // save the total timestep timing
+        funcTimes.push_back((float)numLocalParticles); // save the local number of particles
+    }
+
+    void gatherTimings()
+    {
+        std::cout << "Gathering profiling data." << std::endl;
+
+        float* durations = new float[_numRanks * funcTimes.size()];
```
sekelle: Memory leak here. Why use error-prone `new` here instead of an `std::vector` whose buffer you can access with `data()`?

main/src/util/profiler.hpp
```diff
@@ -0,0 +1,147 @@
...
+        MPI_Gather(funcTimes.data(), funcTimes.size(), MPI_FLOAT, &durations[_rank * funcTimes.size()],
+                   funcTimes.size(), MPI_FLOAT, 0, MPI_COMM_WORLD);
+
+        if (_rank == 0)
+        {
+            for (int i = 0; i < funcTimes.size(); i++)
+            {
+                float* funcDurations = new float[_numRanks];
```
sekelle: same as above

main/src/util/profiler.hpp
```diff
@@ -0,0 +1,147 @@
...
+                      << covPerStep.at(iteration) << "," << I_2PerStep.at(iteration) << ","
+                      << lambdaPerStep.at(iteration);
+        for (int i = 0; i < _numRanks; i++)
+        {
+            profilingFile << "," << vectorMetric.at(iteration)[i];
+        }
+    }
+
+    void printProfilingInfo()
+    {
```
sekelle, line 82: I would only save the raw data, i.e. `funcTimes` plus relevant meta data, e.g. how many ranks, how many measuring points per step and their name string labels. Which means it would make the analysis more robust and easier to deal with and implement if you saved the timing data as HDF5 file attributes.

main/src/propagator/factory.hpp
```diff
@@ -49,27 +49,45 @@ namespace sphexa
...
+    {
+        return std::make_unique<HydroProp<DomainType, ParticleDataType>>(output, rank, isProfilingEnabled);
     }
-    if (choice == "std") { return std::make_unique<HydroProp<DomainType, ParticleDataType>>(output, rank); }
 #ifdef SPH_EXA_HAVE_GRACKLE
     if (choice == "std-cooling")
     {
-        return std::make_unique<HydroGrackleProp<DomainType, ParticleDataType>>(output, rank);
+        return std::make_unique<HydroGrackleProp<DomainType, ParticleDataType>>(output, rank, isProfilingEnabled);
     }
```
sekelle: I believe the `isProfilingEnabled` flags not needed anymore.

PR #376, opened 2023-03-24
Fixed missing initialization in observablesFactory

nknk567:
Missing initialization of khAttribute led to error when using the standard propagator.


lks1248:
I approved #377 with more extensive fixes instead .

sekelle:
Seems like we did exactly the same at exactly the same time.

PR #383, opened 2023-04-14
Protect IAD against numerical underflow

sekelle:


OsmanSeckinSimsek (approved):

sph/include/sph/hydro_std/iad_kern.hpp
```diff
@@ -54,9 +53,9 @@ HOST_DEVICE_FUN inline void IADJLoopSTD(cstone::LocalIndex i, T sincIndex, T K,
     T det = tau11 * tau22 * tau33 + T(2) * tau12 * tau23 * tau13 - tau11 * tau23 * tau23 - tau22 * tau13 * tau13 -
             tau33 * tau12 * tau12;
 
-    // note normalization factor: cij have units of 1/tau because det is proportional to tau^3 so we have to
-    // divide by K/h^3
-    T factor = (hi * hi * hi) / (det * K);
+    // Note normalization factor: cij have units of 1/tau because det is proportional to tau^3 so we have to
+    // divide by K/h^3. det can be zero due to numerical underflow
+    T factor = (det == T(0.0)) ? T(0.0) : (hi * hi * hi) / (det * K);
```
rmcabezon, line 58: I think this is problematic. If det is zero, this will set the IAD coefficients to zero and the error will pass unnoticed, keeping calculations going on but being wrong. If det is zero, it should throw an exception and stop the calculation or give a NaN that can be detected. Please, revert this and correct.

PR #384, opened 2023-04-24
Group targets

sekelle:
When the particle density is not smooth, such as on the surface of an Evrard sphere or any other sharp transitions from material to vacuum, consecutive groups of `GpuConfig::warpSize` particles might contain such a discontinuity and will have a very large bounding box, leading to a huge breadth-first traversal stack.

This change avoids this problem by breaking up these groups into smaller ones.


ryoanji/src/ryoanji/nbody/direct.cuh
```diff
@@ -99,13 +101,14 @@ __global__ void directKernel(int numSource, const T* __restrict__ x, const T* __
 }
 
 template<class T>
-void directSum(std::size_t numBodies, const T* x, const T* y, const T* z, const T* m, const T* h, T* p, T* ax, T* ay,
-               T* az)
+void directSum(size_t first, size_t last, size_t numBodies, const T* x, const T* y, const T* z, const T* m, const T* h,
```
OsmanSeckinSimsek, line 104: isn't numBodies always last-first? maybe we don't need to pass that.

OsmanSeckinSimsek (approved):
approving with a minor comment

ryoanji/src/ryoanji/nbody/direct.cuh
sekelle: For the direct sum we need it. `last - first` is the number of targets, but `numBodies` is the number of sources here, so they're different.

PR #388, opened 2023-05-31
Compute IAD-tau determinants with normalization factors

sekelle:


sph/include/sph/hydro_std/iad_kern.hpp
```diff
@@ -42,27 +45,31 @@ HOST_DEVICE_FUN inline void IADJLoopSTD(cstone::LocalIndex i, T sincIndex, T K,
...
+        tau[I::_12] += rx * ry * mj_roj_w;
+        tau[I::_13] += rx * rz * mj_roj_w;
+        tau[I::_22] += ry * ry * mj_roj_w;
+        tau[I::_23] += ry * rz * mj_roj_w;
+        tau[I::_33] += rz * rz * mj_roj_w;
     }
 
-    T det = tau11 * tau22 * tau33 + T(2) * tau12 * tau23 * tau13 - tau11 * tau23 * tau23 - tau22 * tau13 * tau13 -
-            tau33 * tau12 * tau12;
+    T normalization = T(1) / min(abs(tau));
```
rmcabezon: You need to catch the possibility that one of the taus might be zero. If not this will be NaN.

rmcabezon: Also, if the matrix is very unbalanced, it might happen that one of the taus is smaller than one and in the end, you enlarge the others, which will worsen the problem.

rmcabezon: Why not use as normalization a value calculated at the beginning of the simulation, equal to the largest particle mass times pi times the largest coordinate value?

max_coord = max(max(x), max(y), max(z))
normalization = 1 / (max(m) * pi * max_coord)

I usually add the pi factor because systems may grow beyond their original size (i.e. explosions) and I don't want to make an exact multiple of the size.

sekelle: Yeah that was indeed a silly suggestion on my part. `normalization = 1 / (max(m) * pi * max_coord)` would work well, but we don't have max(m) at the moment. Computing it on the fly would be too expensive and computing it at the start of the simulation requires additional arguments to be passed down to the kernel.

Instead I'm proposing normalizing by the average exponent of the elements of tau. That brings back both small and large `tau`s closer to the origin, or leaves them more or less unchanged should some values be very big and some very small. Also avoids any division by zero problems.

rmcabezon: Your idea of averaging over exponents is good. That would work. But the truth to be told, we probably would never change the mass of the particles and if we do it won't be more than a few times (i.e. x2, x3). What if we simply take the mass of one particle at the beginning and we avoid all calculations?

sekelle: Once we have stars and dark matter, the masses might vary over a wider range compared to the cases that we're currently running. Looking at the average exponents is probably the most robust and also most adaptive.
In terms of cost, it's close to zero. Binary logarithms and raising a power of 2 are cheap.
Perhaps I could revert to the original `tau11`, `tau12` notation if you prefer. That would mean a very small code change. It seems the `tau` as array doesn't help so much anymore with the exponents.

rmcabezon: To me is easier to understand with the tau11, tau12 notation. And it there is almost no cost to the exponents method, let's go for it then.

sekelle: done

sph/include/sph/hydro_std/iad_kern.hpp
```diff
@@ -50,12 +50,24 @@ HOST_DEVICE_FUN inline void IADJLoopSTD(cstone::LocalIndex i, T sincIndex, T K,
         tau33 += rz * rz * mj_roj_w;
     }
 
+    auto getExp    = [](T val) { return (val == T(0) ? 0 : std::ilogb(val)); };
+    int  tauExpSum = getExp(tau11) + getExp(tau12) + getExp(tau13) + getExp(tau22) + getExp(tau23) + getExp(tau33);
```
rmcabezon, line 54: I think this might work. The only problem I can see if that it could happen that the average tauExpSum is zero. In that case, the normalization would be 1 and we might be in the same problem of underflow.

jgphpc (approved):
lgtm

sph/include/sph/hydro_std/iad_kern.hpp
sekelle: If the normalization is 1, then either all exponents are close to zero, or some of them are very big and some very small. In that scenario there's a good chance that the big exponents cancel out with the small ones in the tau products and the underflow scenario would not apply in this case.

PR #390, opened 2023-06-13
Update visualization code & plotting particles with ranks

yzhu-ub:
**This is NOT ready for a merge yet.**

- Updated visualization code for Ascent & Catalyst to adapt to recent API changes
- All visualization scripts have been moved to /scripts/<corresponding-sub-directories>
- Add reference to propagator in viz::execute(...)


CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ if (NOT OpenMP_CXX_INTERFACE_LINK_OPTIONS)
     set_property(TARGET OpenMP::OpenMP_CXX PROPERTY INTERFACE_LINK_OPTIONS ${OpenMP_CXX_INTERFACE_LINK_OPTIONS})
 endif()
 
+include_directories("/home/appcell/spack/opt/spack/linux-ubuntu22.04-skylake/gcc-11.3.0/openmpi-4.1.5-ej4bfoygrfqlhn76um3frnrup3bodmpt/include")
```
sekelle, line 36: I guess this is an oversight. Should be configurable instead of hard-coded.

main/src/ascent_adaptor.h
```diff
@@ -103,41 +104,59 @@ void addField(conduit::Node& mesh, const std::string& name, FieldType* field, si
...
+        conn[i] = i;
+    }
+
+    mesh["topologies/mesh/elements/connectivity"].set_external(conn);
+
+    mesh["fields/particles/association"] = "vertex";
+    mesh["fields/particles/topology"]    = "mesh";
+    mesh["fields/particles/values"].set_external(&d.x[startIndex], endIndex - startIndex);
+
+    std::vector<conduit_int64> ranks(endIndex - startIndex);
```
sekelle, line 140: Here I would use a 32-bit integer, MPI anyway does not support more than 2^31-1 ranks.

sekelle (changes requested):
Thanks for this work, looks really well. Defining a temporary field with the ranks is a good first solution I think.
The only other request I have is to revert all changes that only add or remove an empty line and to squash the commits into a single one. Some top-level directories are shared with other repositories via git-subtrees, therefore it doesn't make much sense to replicate changes there with just empty lines.

jfavre:
FYI, plotting the MPI rank can now be done with a pipeline of type "add_mpi_ranks". Thus, your code could `look` like
```
   conduit::Node pipelines, actions;
    // pipeline 1
    pipelines["pl1/f1/type"] = "add_mpi_ranks";
    conduit::Node &params = pipelines["pl1/f1/params"];
    params["topology"] = "mesh";      
    params["output"] = "ranks"; 

    conduit::Node scenes;
    scenes["s1/plots/p1/type"] = "pseudocolor";
    scenes["s1/plots/p1/field"] = "ranks";
    scenes["s1/plots/p1/pipeline"] = "pl1";
    scenes["s1/plots/p1/color_table/discrete"] = "true";
    scenes["s1/renders/r1/image_prefix"] = "ranks";

    conduit::Node &add_pipelines = actions.append();
    add_pipelines["action"] = "add_pipelines";
    add_pipelines["pipelines"] = pipelines;
    // add the scenes
    conduit::Node &add_scenes= actions.append();
    add_scenes["action"] = "add_scenes";
    add_scenes["scenes"] = scenes;
```

sekelle:
No updates in over a year. There are simpler alternatives are available now for plotting the subdomains, closing.

PR #396, opened 2023-07-20
Change number of particles

nknk567:


domain/include/cstone/domain/domain.hpp
```diff
@@ -321,6 +321,10 @@ class Domain
     [[nodiscard]] LocalIndex startIndex() const { return bufDesc_.start; }
     //! @brief return one past the index of the last particle that's part of the local assignment
     [[nodiscard]] LocalIndex endIndex() const { return bufDesc_.end; }
+    //! @brief set the index of the lsat particle (used to increase the number of particles)
+    void setEndIndex(const size_t i) { bufDesc_.end = i; }
+    //! @brief set the total number of particles (including halos) (used to increase the number of particles)
+    void setBufSize(const size_t i) { bufDesc_.size = i; }
```
sekelle: Resizing the buffer should not be needed, because the size will be automatically inferred in the next call to domain.sync()

sekelle:
needs squashing of commits before merging

PR #399, opened 2023-08-07
Acceleration criterion

nknk567:
Shouldn't this be the acceleration instead of the position?


sekelle (approved):
You're right of course.

PR #400, opened 2023-08-07
Absorbed sinc power into lookup tables

sekelle:
Reference data will need to be slightly adjusted.

### DIFF
--- a/sph/include/sph/hydro_ve/ve_def_gradh_kern.hpp
+++ b/sph/include/sph/hydro_ve/ve_def_gradh_kern.hpp
@@ -68,7 +68,7 @@ veDefGradhJLoop(cstone::LocalIndex i, T sincIndex, T K, const cstone::Box<T>& bo
 
         T dist   = distancePBC(box, hi, xi, yi, zi, x[j], y[j], z[j]);
         T vloc   = dist * hInv;
-        T w      = math::pow(lt::wharmonic_lt_with_derivative(wh, whd, vloc), sincIndex);
+        T w      = lt::wharmonic_lt_with_derivative(wh, whd, vloc);
         T dw     = wharmonic_derivative(vloc, w) * sincIndex;
         T dterh  = -(T(3) * w + vloc * dw);
         T xmassj = xm[j];
--- a/sph/include/sph/hydro_ve/ve_def_gradh_kern.hpp
+++ b/sph/include/sph/hydro_ve/ve_def_gradh_kern.hpp
@@ -42,10 +42,10 @@ namespace sph
 {
 
 template<size_t stride = 1, class Tc, class Tm, class T>
-HOST_DEVICE_FUN inline util::tuple<T, T>
-veDefGradhJLoop(cstone::LocalIndex i, T sincIndex, T K, const cstone::Box<T>& box, const cstone::LocalIndex* neighbors,
-                unsigned neighborsCount, const Tc* x, const Tc* y, const Tc* z, const T* h, const Tm* m, const T* wh,
-                const T* whd, const T* xm)
+HOST_DEVICE_FUN inline util::tuple<T, T> veDefGradhJLoop(cstone::LocalIndex i, T K, const cstone::Box<T>& box,
+                                                         const cstone::LocalIndex* neighbors, unsigned neighborsCount,
+                                                         const Tc* x, const Tc* y, const Tc* z, const T* h, const Tm* m,
+                                                         const T* wh, const T* whd, const T* xm)
 {
     auto xi     = x[i];
     auto yi     = y[i];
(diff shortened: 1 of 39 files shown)


sph/include/sph/hydro_ve/ve_def_gradh_kern.hpp
```diff
@@ -68,8 +68,8 @@ veDefGradhJLoop(cstone::LocalIndex i, T sincIndex, T K, const cstone::Box<T>& bo
 
         T dist   = distancePBC(box, hi, xi, yi, zi, x[j], y[j], z[j]);
         T vloc   = dist * hInv;
-        T w      = math::pow(lt::wharmonic_lt_with_derivative(wh, whd, vloc), sincIndex);
-        T dw     = wharmonic_derivative(vloc, w) * sincIndex;
+        T w      = lt::lookup(wh, vloc);
+        T dw     = lt::lookup(whd, vloc);
         T dterh  = -(T(3) * w + vloc * dw);
```
rmcabezon, line 73: Double-check if this vloc multiplying dw is still valid with the new table. Originally, dW/dv was divided by an additional vloc to avoid the calculation of an extra r. 
Ex: dW/dx=dW/dv*dv/dr*dr/dx=dW/dv*(1/h)*(x/r)=(dW/dv)/v/h^2
That's why this vloc is multiplying in line 73. To compensate for that. I think that the original table had this 1/vloc included. Not sure if this is the same now with the new table.

sph/include/sph/tables.hpp
```diff
@@ -8,45 +8,56 @@ namespace sph
...
+//! @brief create a lookup-table for sinc(x)^sincIndex
 template<typename T, std::size_t N>
-std::array<T, N> createWharmonicLookupTable()
+std::array<T, N> createWharmonicLookupTable(T sincIndex)
 {
+    constexpr int    numIntervals = N - 1;
     std::array<T, N> wh;
 
-    const T halfsSize = N / 2.0;
+    constexpr T dx = 2.0 / numIntervals;
```
rmcabezon, line 20: This 2 is the compact support of the kernel. This can be different for other kernels. I think that for now is ok to not have an additional parameter, but it would be better to declare it as a constant here so that it is clear what it means.

sph/include/sph/tables.hpp
```diff
@@ -8,45 +8,56 @@ namespace sph
...
     for (size_t i = 0; i < N; ++i)
     {
-        T normalizedVal = i / halfsSize;
-        wh[i]           = wharmonic_std(normalizedVal);
+        T normalizedVal = i * dx;
```
rmcabezon, line 23: Does C++ lose precision when multiplying an int with a double? Shouldn't the i be casted as double?

sph/include/sph/tables.hpp
```diff
@@ -8,45 +8,56 @@ namespace sph
...
+//! @brief create a lookup-table for d(sinc(x)^sincIndex)/dx
 template<typename T, std::size_t N>
-std::array<T, N> createWharmonicDerivativeLookupTable()
+std::array<T, N> createWharmonicDerivativeLookupTable(T sincIndex)
 {
+    constexpr int    numIntervals = N - 1;
     std::array<T, N> whd;
 
-    const T halfsSize = N / 2.0;
+    const T dx = 2.0 / numIntervals;
```
rmcabezon, line 36: Same here as before regarding the compact support.

sph/include/sph/tables.hpp
```diff
@@ -8,45 +8,56 @@ namespace sph
...
     for (size_t i = 0; i < N; ++i)
     {
-        T normalizedVal = i / halfsSize;
-        whd[i]          = wharmonic_derivative_std(normalizedVal);
+        T normalizedVal = i * dx;
```
rmcabezon, line 39: Same here regarding precision.

rmcabezon (commented):
Is there still a way to use a non-interpolated kernel by the user's choice?

sph/include/sph/hydro_ve/ve_def_gradh_kern.hpp
sekelle: `dw` is still just the plain kernel derivative at `vloc`, nothing changed in that regard. This is covered by the kernel test for the `gradh` value which will come out wrong if `dw` is changed.

sph/include/sph/tables.hpp
sekelle: C++ will do that automatically (at compile time).

sph/include/sph/tables.hpp
sekelle: I already introduced naming the support explicitly in the `lookup` function which is general. This one here will be changed and made more general when you introduce the new weighted sinc sum kernel.

sekelle:
> Is there still a way to use a non-interpolated kernel by the user's choice?

If you replace the lookup by the direct call to the kernel, yes. This PR doesn't change the flexibility in that regard, it only increases the flexibility of what kernel you can use for the table.

rmcabezon (approved):

PR #401, opened 2023-08-13
Update cooling

nknk567:
Update for the radiative cooling.
Changes made:
- Propagator now also calculates a cooling timestep
- Propagator calculates pressure via grackle (important if H2 is present)
- Function loadOrStoreAttributes() implemented in cooler.hpp allowing to read and write Cooling parameters to file. This happens in load and save functions of the propagator.
- Adapted Evrard init and cooling tests for the new layout.
- Using internal energy u instead of temperature if cooling is activated (otherwise tedious conversions are required).

### DIFF
--- a/extern/grackle/CMakeLists.txt
+++ b/extern/grackle/CMakeLists.txt
@@ -37,7 +37,7 @@ set(grackle_SRCS
 add_library(grackle ${grackle_SRCS})
 
 set(PRECISION_DEFINE GRACKLE_FLOAT_8)
-configure_file(${PROJECT_SOURCE_DIR}/config/grackle_float.in.h ${CMAKE_BINARY_DIR}/extern/grackle/grackle_deps/grackle_float.h)
+configure_file(${PROJECT_SOURCE_DIR}/config/grackle_float.h.in ${CMAKE_BINARY_DIR}/extern/grackle/grackle_deps/grackle_float.h)
 
 target_include_directories(grackle PUBLIC ${PROJECT_SOURCE_DIR}/grackle_repo/src/clib)
 target_include_directories(grackle PUBLIC ${PROJECT_BINARY_DIR}/grackle_deps)
--- a/extern/grackle/CMakeLists.txt
+++ b/extern/grackle/CMakeLists.txt
@@ -4,47 +4,53 @@ project(grackle C Fortran)
 find_package(HDF5 REQUIRED)
 
 set(grackle_SRCS
-    grackle_deps/auto_show_config.c
-    grackle_deps/auto_show_flags.c
-    grackle_deps/auto_show_version.c
-    grackle_repo/src/clib/calculate_cooling_time.c
-    grackle_repo/src/clib/calculate_dust_temperature.c
-    grackle_repo/src/clib/calculate_gamma.c
-    grackle_repo/src/clib/calculate_pressure.c
-    grackle_repo/src/clib/calculate_temperature.c
-    grackle_repo/src/clib/grackle_units.c
+        grackle_deps/auto_show_config.c
+        grackle_deps/auto_show_flags.c
+        grackle_deps/auto_show_version.c
+        grackle_repo/src/clib/calculate_cooling_time.c
+        grackle_repo/src/clib/calculate_dust_temperature.c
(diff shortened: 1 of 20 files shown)


extern/grackle/CMakeLists.txt
```diff
@@ -35,11 +35,17 @@ set(grackle_SRCS
...
+configure_file(${PROJECT_SOURCE_DIR}/config/grackle_float.in.h ${CMAKE_BINARY_DIR}/extern/grackle/grackle_deps/grackle_float.h)
+
 target_include_directories(grackle PUBLIC ${PROJECT_SOURCE_DIR}/grackle_repo/src/clib)
-target_include_directories(grackle PUBLIC ${CMAKE_BINARY_DIR}/extern/grackle)
+target_include_directories(grackle PUBLIC ${PROJECT_BINARY_DIR}/grackle_deps)
+target_include_directories(grackle PRIVATE ${HDF5_INCLUDE_DIRS} ${MPI_C_INCLUDE_PATH})
 target_link_libraries(grackle ${HDF5_LIBRARIES} ${MPI_C_LIBRARIES} gfortran)
-target_compile_definitions(grackle PRIVATE LINUX H5_USE_16_API CONFIG_BFLOAT_8 PIC)
+#Omit warning of deprecated functions (already from removed SPH-EXA)
+target_compile_definitions(grackle PUBLIC OMIT_LEGACY_INTERNAL_GRACKLE_FUNC)
```
sekelle, line 47: This compile definition should probably be `PRIVATE`

physics/cooling/CMakeLists.txt
```diff
@@ -4,11 +4,12 @@ project(sphexa-cooling)
 set(CMAKE_CXX_STANDARD 20)
 
 set(CSTONE_DIR ${CMAKE_SOURCE_DIR}/domain/include)
+set(FILE_DIR ${CMAKE_SOURCE_DIR}/main/src)
```
sekelle: `MAIN_APP_DIR` would be a better name

extern/grackle/CMakeLists.txt
nknk567: Problem is that the warnings are generated in grackle.h which is included by SPH-EXA.

physics/cooling/CMakeLists.txt
nknk567: I will change it.

main/src/propagator/std_hydro_grackle.hpp
```diff
@@ -64,7 +65,7 @@ class HydroGrackleProp final : public HydroProp<DomainType, DataType>
      *
      * x, y, z, h and m are automatically considered conserved and must not be specified in this list
      */
-    using ConservedFields = FieldList<"temp", "vx", "vy", "vz", "x_m1", "y_m1", "z_m1", "du_m1">;
+    using ConservedFields = FieldList<"u", "vx", "vy", "vz", "x_m1", "y_m1", "z_m1", "du_m1", "alpha">;
```
sekelle: I don't think you're using the "alpha" field.

nknk567: True, I removed it

sph/include/sph/find_neighbors.hpp
```diff
@@ -28,7 +28,9 @@ void findNeighborsSph(const T* x, const T* y, const T* z, T* h, LocalIndex first
         int iteration = 0;
         while (ngmin > nc[i] || nc[i] > ngmax && iteration++ < maxIteration)
         {
-            h[id] = updateH(ng0, nc[i], h[id]);
+            if (nc[i] == 0)
+                h[id] *= 1.25;
+            else { h[id] = updateH(ng0, nc[i], h[id]); }
```
sekelle: `nc[i]` does not include the self reference. I think here it would be permissible to use `nc[i] + 1`  in `updateH` and avoid the division by zero problem. This would also address it on the GPU while this "fix" here would only be used on the CPU.

sekelle: Addressed here
https://github.com/unibas-dmi-hpc/SPH-EXA/pull/406

nknk567:
new PR from main repository

PR #407, opened 2023-08-28
use cub double buffer radix sort

sekelle:
GPU memory optimizations:
* Replaced Thrust radix sort with lower-level CUB interface that avoids allocation of large temporary arrays by reusing existing buffers
* Separation of GPU buffers used for send/receive staging and GPU buffers used for relocation+swap operations to avoid cyclic swap chain

Performance optimizations:
* Added support for hydro fields in FP32 with only coordinates and temperatur/internal energy stored in FP64.

### DIFF
--- a/sph/include/sph/tables.hpp
+++ b/sph/include/sph/tables.hpp
@@ -12,7 +12,7 @@ constexpr int size = 20000;
 
 //! @brief create a lookup-table for sinc(x)^sincIndex
 template<typename T, std::size_t N>
-std::array<T, N> createWharmonicLookupTable(T sincIndex)
+std::array<T, N> createWharmonicTable(double sincIndex)
 {
     constexpr int    numIntervals = N - 1;
     std::array<T, N> wh;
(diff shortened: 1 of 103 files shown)


sph/include/sph/tables.hpp
```diff
@@ -12,7 +12,7 @@ constexpr int size = 20000;
 
 //! @brief create a lookup-table for sinc(x)^sincIndex
 template<typename T, std::size_t N>
-std::array<T, N> createWharmonicLookupTable(T sincIndex)
+std::array<T, N> createWharmonicTable(double sincIndex)
```
OsmanSeckinSimsek, line 15: not sure I understand the benefit of double for sincIndex here.

OsmanSeckinSimsek (approved):

sph/include/sph/tables.hpp
sekelle: With `T sincIndex` there's type-deduction of the table based on the type of `sincIndex` which is not what is needed.
Performance is not an issue here, because it's just the construction of the lookup table.

PR #412, opened 2023-09-19
Fix io-bug and clean-up and cmake configs

sekelle:
Bug was introduced in chem-write


nknk567 (approved):
Tested with two steps of Evrard

PR #416, opened 2023-10-08
 Kernel for calculating additional fields that require neigbor traversal

lks1248:
initial implementation, including `markRamp` calculation. I would be happy to take some feedback on the general layout etc. I will test the functionality and correctness before moving the PR to ready for merging.


lks1248:
The calculations work fine,  I am just wondering if I could deallocate `markRamp` most of the time, it seems to be allocated by default.

PR #419, opened 2023-10-25
Abstract file I/O

sekelle:
* perform file I/O consistently through abstract interfaces
* allow test-case specific settings to be overridden with a settings file containing HDF5 attributes

### DIFF
--- a/main/src/io/ifile_io_hdf5.hpp
+++ b/main/src/io/ifile_io_hdf5.hpp
@@ -23,7 +23,7 @@
  */
 
 /*! @file
- * @brief file I/O interface
+ * @brief File I/O interface implementation with H5Part
  *
  * @author Sebastian Keller <sebastian.f.keller@gmail.com>
  */
@@ -33,7 +33,6 @@
 #include <mpi.h>
 
 #include <filesystem>
-#include <map>
 #include <string>
 #include <variant>
 #include <vector>
@@ -41,13 +40,13 @@
 #include "init/grid.hpp"
 
 #include "file_utils.hpp"
-#include "mpi_file_utils.hpp"
+#include "h5part_wrapper.hpp"
 #include "ifile_io.hpp"
 
 namespace sphexa
 {
 
-class H5PartWriter : public IFileWriter
+class H5PartWriter final : public IFileWriter
 {
 public:
     using Base      = IFileWriter;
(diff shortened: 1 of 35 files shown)


scripts/set_parms.py
```diff
@@ -0,0 +1,61 @@
...
+
+"""
+Command line utility to create an HDF5 simulation parameters file with key-value pairs
+
+Usage examples:
+    $ settings.py <parameter_file> [--add] [list of settings as key-value-pairs, e.g --ng0 100]'
+"""
+
+__program__ = "set_parms.py"
+__author__ = "Sebastian Kelelr (sebastian.f.keller@gmail.com)"
```
OsmanSeckinSimsek: typo in the name

sph/include/sph/hydro_turb/turbulence_data.hpp
```diff
@@ -55,54 +55,61 @@ class TurbulenceData
...
     template<class Archive>
     void loadOrStore(Archive* ar)
     {
-        ar->stepAttribute("turbulencePhases", phases.data(), phases.size());
+        std::string prefix = "turbulence::";
+        ar->stepAttribute(prefix + "variance", &variance, 1);
+        ar->stepAttribute(prefix + "decayTime", &decayTime, 1);
+        ar->stepAttribute(prefix + "solWeight", &solWeight, 1);
+        ar->stepAttribute(prefix + "solWeightNorm", &solWeightNorm, 1);
+        ar->stepAttribute(prefix + "numModes", &numModes, 1);
```
OsmanSeckinSimsek, line 97: This line gives an error while compiling locally on Clang 15.0.0 on Mac:
error: no viable conversion from 'size_t *' (aka 'unsigned long *') to 'FieldType' (aka 'variant<double *, float *, char *, int *, long long *, unsigned int *, unsigned long long *>')
ar->stepAttribute(prefix + "numModes", &numModes, 1);

maybe use uint64_t numModes;

main/src/io/ifile_io_hdf5.hpp
```diff
@@ -120,18 +106,26 @@ class H5PartWriter : public IFileWriter
...
+        }
+    }
 
 private:
+    int      rank_{0};
     MPI_Comm comm_;
 
-    size_t      firstIndex_, lastIndex_;
+    size_t      firstIndex_{0}, lastIndex_{0};
     std::string pathStep_;
```
OsmanSeckinSimsek, line 122: variable not used.

main/test/mpi/hdf5.cpp
OsmanSeckinSimsek, line 1: in the test turbulenceData line 246 : {"powerLawExp", 5 / 3} -> {"powerLawExp", 5. / 3.}

OsmanSeckinSimsek (changes requested):
There are a couple of minor changes that is only suggestions such as unused variable.
The important change request is the hdf5.cpp file TurbulenceData test powerLawExp variable being 5/3.

main/src/io/ifile_io_hdf5.hpp
sekelle: removed

scripts/set_parms.py
sekelle: fixed

sph/include/sph/hydro_turb/turbulence_data.hpp
sekelle: yeah, uint64_t is better, I changed that

main/test/mpi/hdf5.cpp
sekelle: It doesn't really matter since this is just a store / load test, but I took the opportunity to remove all unneeded settings from that map.

OsmanSeckinSimsek (approved):

PR #421, opened 2023-10-29
Smoothing kernel selection at runtime

sekelle:
Adds a numerical integrator to determine the normalization constant `K`

`kernelChoice` is now a recognized setting that can be used to select at runtime between different smoothing kernels.
The tables and normalization constant are handled automatically.

Example usage:

```bash
$ ~/soft/SPH-EXA/scripts/set_parms.py ktype1.h5 --kernelChoice 1
ktype1.h5 now contains the following settings:
   kernelChoice 1
$ ./sphexa-cuda --init sedov:ktype1.h5 --glass 50c.h5 -s 200
```

```bash
# SPHEXA: integrator-1d/8f6bd1c1
# 1 MPI-4.0 process(es) with 1 OpenMP-201511 thread(s)/process
Override setting from ktype1.h5: kernelChoice = 1
Data generated for 125000 global particles
Domain synchronized, nLocalParticles 125000
# domain::sync: 0.00177403s
```


sph/include/sph/kernels.hpp
rmcabezon, line 1: Did you completely remove the analytical Harmonic kernel?

sph/include/sph/table_creation.hpp
```diff
@@ -0,0 +1,163 @@
...
+    return b0 + b1 * std::sqrt(n) + b2 * n + b3 * std::sqrt(n * n * n);
+}
+
+//! @brief compute the 3D normalization constant for an arbitrary kernel
+template<class F>
+double kernel_3D_k(F&& sphKernel, double support)
+{
+    auto kernelVol3D = [sphKernel](double x) { return 4.0 * M_PI * x * x * sphKernel(x); };
+
+    uint64_t numIntervals = 2000;
```
rmcabezon: This number of intervals hardcoded might not be enough for some kernels if they are very steep. The integration method that you use has a way to evaluate the error. Should we add something like that to decide the numIntervals?

sph/include/sph/table_creation.hpp
rmcabezon, line 1: table_creation is too ambiguous for what this is. The word "kernel" should be included in the name of this file. Maybe "SPHkernel_table_creation"? Otherwise, I surely will forget where this is done.

sph/include/sph/kernels.hpp
sekelle: No, it's still there in `kernels.hpp`

sph/include/sph/table_creation.hpp
sekelle: It's already overkill as it is, as long as the smoothing kernel is smooth. The steepness (=first derivative) only affects  the accuracy of Riemann sums. Simpsons rule is unaffected by the steepness or curvature, because it handles anything up to cubic order exactly. That's why  the error estimation that you mentioned is proportional to the fourth derivative. That would be cumbersome to evaluate in practice. We also hardcode the number of interpolation points for the lookup table, which will become a problem much sooner than this numerical integration.

sph/include/sph/table_creation.hpp
sekelle: Sure I can rename that, e.g. to `sph_kernel_tables.hpp`.

rmcabezon (approved):

PR #422, opened 2023-11-01
Add time and energy measurements

OsmanSeckinSimsek:
This commit adds time and energy measurements for each function.
Energy is measured using PMT on Cray systems if enabled.

### DIFF
--- a/domain/include/cstone/sfc/hilbert.hpp
+++ b/domain/include/cstone/sfc/hilbert.hpp
@@ -116,8 +116,7 @@ iHilbert(unsigned px, unsigned py, unsigned pz) noexcept
  */
 
 template<class KeyType>
-constexpr HOST_DEVICE_FUN inline std::enable_if_t<std::is_unsigned_v<KeyType>, KeyType> iHilbert2D(unsigned px,
-                                                                                                   unsigned py) noexcept
+HOST_DEVICE_FUN std::enable_if_t<std::is_unsigned_v<KeyType>, KeyType> iHilbert2D(unsigned px, unsigned py) noexcept
 {
     assert(px < (1u << maxTreeLevel<KeyType>{}));
     assert(py < (1u << maxTreeLevel<KeyType>{}));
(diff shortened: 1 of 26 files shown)


main/src/util/profiler.hpp
```diff
@@ -0,0 +1,320 @@
...
+
+    void printProfilingInfoHDF5(size_t timesteps)
+    {
+        H5PartSetStep(_profilingFile, 0);
+        H5PartSetNumParticles(_profilingFile, funcTimes.size());
+        int numFuncs = funcTimes.size() / timesteps;
+        fileutils::writeH5PartStepAttrib(_profilingFile, "NUM_RANKS", &_numRanks, 1);
+        fileutils::writeH5PartStepAttrib(_profilingFile, "NUM_FUNCS", &numFuncs, 1);
+        std::string fieldName = "RANK" + std::to_string(_rank);
+        fileutils::writeH5PartField(_profilingFile, fieldName, funcTimes.data());
```
sekelle: This will give you a random label. For parallel writes, the field name has to match across ranks.

sekelle (changes requested):
Thanks for the code. The biggest opportunity to improve this PR is by reusing the `IFileWriter` interface. This will a) get rid of separate HDF5 and Ascii write functions and b) avoid MPI gather code because `IFileWriter` takes care of that. This would be good news for modularization to cut down on compile times as well because we can compile the IO stack in separate unit.
A second factor to consider is to perform the energy measurements through an abstract interface. You shouldn't have to #ifdef every single function in the CrayPM reader and you might at some point want to get readings from a different provider.

main/src/util/profiler.hpp
OsmanSeckinSimsek: I thought HDF5 could write into multiple labels in parallel. Is it not the case?

sekelle:
I have reimplemented this for the following reasons:
* Counter reading (`pmt.h`) was missing
* Perform IO only through the abstract interface to eliminate the need for MPI code and separate HDF5/ASCII output.
* Do not hard-code number of accelerators per node, read SLURM env variables when available
* Allow file output frequency to be adjusted instead of writing recorded data only at the end of a simulation

PR #423, opened 2023-11-01
Update README.md

rmcabezon:
Updated information about compiler and modules to compile at sciCORE.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,9 @@ SPH-EXA
 ```
 #### Toolchain requirements
 
-The C++ (.cpp) part of the code requires a **C++20 compiler**, at least GCC 10, clang 12 or cray-clang 14.
+The C++ (.cpp) part of the code requires a **C++20 compiler**, at least GCC 11, clang 12 or cray-clang 14.
 For CUDA (.cu), the minimum supported CUDA version is **CUDA 11.2** with a C++17 host compiler, e.g. GCC 9.3.0.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,10 +63,9 @@ SPH-EXA
 ```
 #### Toolchain requirements
 
-The C++ (.cpp) part of the code requires a **C++20 compiler**, at least GCC 10, clang 12 or cray-clang 14.
+The C++ (.cpp) part of the code requires a **C++20 compiler**, at least GCC 11, clang 12 or cray-clang 14.
 For CUDA (.cu), the minimum supported CUDA version is **CUDA 11.2** with a C++17 host compiler, e.g. GCC 9.3.0.
 
-Note that GCC 10.3.0 does not work as CUDA host compiler due to known compiler bugs.
 For ease of use, the recommended minimum version of CUDA is 11.4.1 which supports GCC 11, providing both the required
 C++20 support and bug-free CUDA host compilation. [**NOTE:** CUDA/11.3.1 seems to have solved the compatibility issues with GCC 10.3.0]
```
sekelle, line 70: The note about CUDA/11.3.1 and GCC 10.3 can be removed now.

sekelle (approved):

README.md
rmcabezon: I thought I removed it in the PR already.

PR #428, opened 2023-12-03
Generate fixed boundary particle layer from mirroring simulation data

lks1248:
Supports flexible width with storing/reloading


domain/include/cstone/sfc/box.hpp
```diff
@@ -113,11 +113,12 @@ class Box
 {
 
 public:
-    HOST_DEVICE_FUN constexpr Box(T xyzMin, T xyzMax, BoundaryType b = BoundaryType::open)
```
sekelle, line 116: The FBC thickness is a detail of your current implementation of fixed boundaries. It should not be stored in the box or be part of a public interface.

sekelle (changes requested):
I remain optimistic that there can be a combination of conservative, repulsive and dissipative forces to make the boundary fixed. This would be easier and cleaner to implement and it would be much simpler to test and validate.

PR #431, opened 2023-12-19
Implements the Sod Shock test case

lks1248:
This depends on the fixed boundary implementation #516


lks1248:
The test case should be re-run and re-verified given that it has been a while. @rmcabezon

PR #432, opened 2023-12-25
Add ADIOS2 integration

yzhu-ub:
- Integrate ADIOS2 alongside H5Part for checkpoint I/O
- Add optional compression settings into command-line interface (including: compressor, params)
- Separate initCond and glassBlock file readers

Todo:
- Add I/O tests


CMakeLists.txt
```diff
@@ -72,6 +72,26 @@ if (SPH_EXA_WITH_H5PART)
     endif ()
 endif ()
 
+# By default, ADIOS2 should be disabled
+option(SPH_EXA_WITH_ADIOS "Enable ADIOS2 IO" OFF)
+if (SPH_EXA_WITH_ADIOS)
+    set(adios_components C)
+    find_package(MPI COMPONENTS ${adios_components})
```
sekelle: We already have `find_package(MPI REQUIRED)` above.

CMakeLists.txt
```diff
@@ -72,6 +72,26 @@ if (SPH_EXA_WITH_H5PART)
...
+    if(MPI_FOUND)
+        # Workaround for various MPI implementations forcing the link of C++ bindings
+        add_definitions(-DOMPI_SKIP_MPICXX -DMPICH_SKIP_MPICXX)
+        list(APPEND adios_components MPI)
```
sekelle, line 83: what happens if C++ libraries are linked?

domain/include/cstone/domain/domain.hpp
```diff
@@ -648,4 +648,4 @@ class Domain
     std::vector<KeyType> swapKeys_;
 };
 
-} // namespace cstone
+} // namespace cstone
```
sekelle: While not a strict requirement since C++11 anymore, we're still adhering to the convention of trailing new-lines in source files. Could you please undo these changes?

main/src/init/settings.hpp
```diff
@@ -44,11 +44,13 @@ using InitSettings = std::map<std::string, double>;
...
 inline void writeSettings(const InitSettings& settings, const std::string& path, IFileWriter* writer)
 {
-    if (std::filesystem::exists(path))
-    {
-        throw std::runtime_error("Cannot write settings: file " + path + " already exists\n");
-    }
+    // if (std::filesystem::exists(path))
+    // {
+    //     throw std::runtime_error("Cannot write settings: file " + path + " already exists\n");
+    // }
```
sekelle, line 50: This check is there not because it's not possible to continue. It prevents output to an existing file. It's true that H5Part, for example, prevents you from overwriting an existing step, so it _should_ be safe to add additional steps to an existing file. If we're 100% confident that file corruption of existing steps is not possible even in the event of a crash, we can discuss allowing this.

main/src/io/factory.hpp
```diff
@@ -31,18 +31,51 @@
...
 
-std::unique_ptr<IFileReader> fileReaderFactory(bool /*ascii*/, MPI_Comm comm) { return makeH5PartReader(comm); }
+std::unique_ptr<IFileReader> fileReaderFactory(bool /*ascii*/, MPI_Comm comm, const std::string& filePath = "",
+                                               const std::string& compression = "")
+{
+    auto suffix = std::filesystem::path(filePath).extension().string();
+    if (suffix == ".bp")
+    {
+#ifdef SPH_EXA_HAVE_ADIOS
+        return makeADIOSReader(comm, compression);
```
sekelle, line 74: I think the compression settings should not be required to initalize `AdiosReader`, they should be stored in the checkpoint file if ADIOS is not already doing that. Otherwise you won't be able to read a file if you lost what settings were used.

If you have the settings available in the file, the `AdiosReader`, should be able to read any file, also uncompressed HDF5, saving you the trouble of passing two file readers to all the simulation initializers.

sekelle (changes requested):
I'm excited to try this out. Unfortunately I couldn't compile this with ADIOS 2.9.2 on Daint:
```bash
SPH-EXA/main/src/io/adios_wrapper.hpp:196:33: error: no matching function for call to 'adios2::Variable<long unsigned int>::AddOperation(std::map<std::__cxx11::basic_string<char>, adios2::Operator>::mapped_type&, <brace-enclosed initializer list>)'
```
Did you test this branch on Piz Daint?

yzhu-ub:
> I'm excited to try this out. Unfortunately I couldn't compile this with ADIOS 2.9.2 on Daint:
> 
> ```shell
> SPH-EXA/main/src/io/adios_wrapper.hpp:196:33: error: no matching function for call to 'adios2::Variable<long unsigned int>::AddOperation(std::map<std::__cxx11::basic_string<char>, adios2::Operator>::mapped_type&, <brace-enclosed initializer list>)'
> ```
> 
> Did you test this branch on Piz Daint?

The issue you have mentioned here is fixed. Tests on Piz Daint are still ongoing, but due to a different issue (tADIOS2 not being correctly linked with several compressors I'm currently testing). I would suggest postponing the review of this PR until I complete the documentation for compiling ADIOS2 on Piz Daint in another README

CMakeLists.txt
yzhu-ub: That's a good question -- I copied this and the next one (https://github.com/unibas-dmi-hpc/SPH-EXA/pull/432#discussion_r1438855751) from ADIOS2 documentation, without actually testing them thoroughly. I will come back with more details on consequences.

main/src/init/settings.hpp
yzhu-ub: Sorry! I didn't realize I've commented these lines (during tests) until you pointed out. Should be fixed.

main/src/io/factory.hpp
yzhu-ub: That's a very good consideration and will involve some additional work. I'll leave it unresolved for the time being and revisit it after resolving the Piz Daint issue. In the meantime, to maintain simplicity in the modifications, I propose that we minimize the changes to the I/O APIs in develop branch until then.

PR #434, opened 2024-01-18
Execute Grackle cooling in OMP blocks. Convert Xmass to density in computeDensity instead of in eos.

nknk567:


sekelle:
I added a missing OpenMP loop and replaced a cuda kernel with thrust::transform.
Other than that I noticed the following points:

* Interface duplication: Your new interface is more powerful and flexible. Passing `numParticles=1` corresponds to the old interface to cool a single particle. Is there any reason to keep the old interface instead of e.g. only exposing
```c++
void cool_particles(T dt, const T* rho, const T* u, T* du, const ParticleType& firstParticle, std::size_t numParticles);
```

This would also allow you to get rid of the old `struct cooler_field_data_content` and you wouldn't need a nearly duplicate test. It's a fair amount of non-trivially maintainable code, so these simplifications would be non-negligible.

* The tasking logic is implemented in the facade and thus exposed to the rest of the code, but could/should be internal to the your GRACKLE frontend.

physics/cooling/include/cooling/cooler_impl.hpp
```diff
@@ -0,0 +1,389 @@
...
+            std::array<T, BlockSize> rho_copy;
+            std::array<T, BlockSize> u_copy;
+
+            const block b(i, t);
+            copyToBlock(rho, rho_copy, b);
+            copyToBlock(u, u_copy, b);
+
+            //forEachTransform(util::get<Fractions, CoolingFields>(particle), rho, b.first, b.first + b.len,
+            //                 std::multiplies<>{});
+            convertFracToDens(particle, rho, b);
```
sekelle: I can "forget" to do the conversion from fractions to densities, either here or below by commenting the conversion, and the grackle cooling tests still pass!

sekelle (changes requested):

physics/cooling/include/cooling/cooler_impl.hpp
nknk567: Rho was = 1 in the test. Test was updated.

sekelle (approved):

PR #435, opened 2024-02-14
Ewald on GPUs

sekelle:
Ewald gravity is now automatically called whenever the bounding box is periodic and gravity is activated.4


jpcoles (approved):
Looks good!

PR #441, opened 2024-04-17
Density probability distribution calculation

lks1248:
Adds a parallel implementation to calculate the Density probability distribution function for the turbulence test


main/src/analytical_solutions/turbulence/density_pdf.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include "density_pdf.hpp"
+#include "io/arg_parser.hpp"
+#include "io/factory.hpp"
+#include "util/utils.hpp"
+#include "cstone/primitives/mpi_wrappers.hpp"
+
+using namespace sphexa;
+using T = float;
+
+int main(int argc, char** argv)
```
rmcabezon, line 39: Maybe add some documentation regarding these arguments and their default values (f.e. bib reference from where they are taken from).

main/src/analytical_solutions/turbulence/density_pdf.cpp
```diff
@@ -0,0 +1,111 @@
...
+    referenceDensity /= globalNumParticles;
+
+    bins = computeProbabilityDistribution(rho, referenceDensity, nBins, minValue, maxValue);
+    std::vector<T> reduced_bins(nBins, 0.0);
+    MPI_Reduce(bins.data(), reduced_bins.data(), nBins, MpiType<T>{}, MPI_SUM, 0, MPI_COMM_WORLD);
+
+    if (rank == 0)
+    {
+        std::for_each(reduced_bins.begin(), reduced_bins.end(),
+                      [globalNumParticles](T& i) { i /= globalNumParticles; });
```
rmcabezon: I think that the normalization factor should be the total number of data points * the width of the bin: globalNumParticles * binSize, because when you calculate the integral of the PDF you do sum(binvalue * binSize), so you have to divide by binSize in the normalization so that the area is 1.

lks1248: done!

main/src/analytical_solutions/turbulence/density_pdf.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+    bins = computeProbabilityDistribution(rho, referenceDensity, nBins, minValue, maxValue);
+    std::vector<T> reduced_bins(nBins, 0.0);
+    MPI_Reduce(bins.data(), reduced_bins.data(), nBins, MpiType<T>{}, MPI_SUM, 0, MPI_COMM_WORLD);
+
+    if (rank == 0)
+    {
+        T binSize = (maxValue - minValue) / nBins;
+        std::for_each(reduced_bins.begin(), reduced_bins.end(),
+                      [globalNumParticles, binSize](T& i) { i /= globalNumParticles * binSize; });
```
rmcabezon: Is this actually dividing by binSize? Or multiplying? I would say that a parethesis is missing, but I'm not sure how the /= operator works.

lks1248: yes, the `/=` Operator divides the left hand side by the right hand side, so it evaluates the product first.

lks1248:
This version doesn't throw an error if it cannot load all particles, I need to look at that before merging.

main/src/analytical_solutions/turbulence/density_pdf.cpp
```diff
@@ -0,0 +1,137 @@
...
+    rho.shrink_to_fit();
+    if (rho.size() != localNumParticles)
+    {
+        throw std::runtime_error("rho length doesn't match local count: " + std::to_string(rho.size()) + "\t" +
+                                 std::to_string(localNumParticles));
+    }
+
+    h5reader->closeStep();
+
+    T localTotalDensity = 0.0;
```
sekelle: This variable should be `double` to avoid loss of precision. (`rho` in `float` is fine)

sekelle: But: why do you even compute this, instead of just dividing the total mass by the volume of the simulation box?

PR #442, opened 2024-04-24
Update add_m1.py

rmcabezon:
The original script was adding the fields to the last step, which are sorted alphabetically, not numerically, in the HDF5, which led to the problem that the fields were added to wrong step. There were also warnings regarding attempts to use a numpy array in a context that requires a scalar. Specifically with minDt. This is solved, too. Finally, the mass of the particles is also added. It is calculated directly as 1./num_particles, which is correct for some scenarios, but not all.

### DIFF
--- a/scripts/add_m1.py
+++ b/scripts/add_m1.py
@@ -2,28 +2,68 @@ import h5py
 import numpy as np
 import sys
 
-# usage: python add_m1.py <hdf5-output-file>
-
 inputFilename = sys.argv[1]
 h5File = h5py.File(inputFilename, "r+")
 
-latestH5Step = list(h5File["/"])[-1]
+
+# Get all step names from the HDF5 file
+step_names = list(h5File["/"])
+# Extract step numbers, convert to integers, and sort
+step_numbers = [int(name.split('#')[1]) for name in step_names]
+step_numbers.sort()
+
+# Finding the latest step by selecting the maximum step number
+latest_step_number = step_numbers[-1]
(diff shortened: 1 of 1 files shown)


scripts/add_m1.py
```diff
@@ -2,28 +2,68 @@
...
+# Get the corresponding group in the HDF5 file
 stepHandle = h5File[latestH5Step]
+print(f"Working on the latest step: {latestH5Step}")
+
+num_particles = stepHandle.attrs["numParticlesGlobal"]  # Get the number of particles
+
+if isinstance(num_particles, np.ndarray):
+    num_particles = num_particles.item()  # Convert to integer if it's an array
+
+num_particles = int(num_particles)  # Convert float to integer
```
sekelle, line 28:
```suggestion
num_particles = stepHandle.attrs["numParticlesGlobal"][0]
```
You can retrieve the total number of particles like this. It's always a numpy array with a 64-bit integer data type. No type check or conversion required.

scripts/add_m1.py
```diff
@@ -2,28 +2,68 @@
...
+
+# Define the constant mass value for each particle
+mass_per_particle = 1. / float(num_particles)
```
sekelle, line 31:
```suggestion
mass_per_particle = 1 / num_particles
```
Integer division is automatically done in floating point in Python.

sekelle (approved):
Looks good to me. Some minor simplifications possible.

PR #452, opened 2024-08-27
ParticleID

OsmanSeckinSimsek:
Initialise the particle IDs at the beginning of the simulation. the id field is conserved throughout the simulation, but does not change during the execution.

### DIFF
--- a/main/src/init/evrard_init.hpp
+++ b/main/src/init/evrard_init.hpp
@@ -71,6 +71,8 @@ void initEvrardFields(Dataset& d, const std::map<std::string, double>& constants
     std::fill(d.y_m1.begin(), d.y_m1.end(), 0.0);
     std::fill(d.z_m1.begin(), d.z_m1.end(), 0.0);
 
+    generateParticleIDs(d.id);
+
     auto cv    = sph::idealGasCv(d.muiConst, d.gamma);
     auto temp0 = constants.at("u0") / cv;
     std::fill(d.temp.begin(), d.temp.end(), temp0);
@@ -178,8 +180,6 @@ public:
         d.loadOrStoreAttributes(&attributeSetter);
 
         initEvrardFields(d, settings_);
-
-        generateParticleIDs(d, rank, numRanks);
 
         return globalBox;
     }
--- a/main/src/init/evrard_init.hpp
+++ b/main/src/init/evrard_init.hpp
@@ -179,6 +179,8 @@ public:
 
         initEvrardFields(d, settings_);
 
+        generateParticleIDs(d, rank, numRanks);
+
         return globalBox;
     }
 
--- a/main/src/init/utils.hpp
+++ b/main/src/init/utils.hpp
@@ -120,6 +120,29 @@ void readFileAttributes(InitSettings& settings, const std::string& settingsFile,
     }
 }
 
+//! @brief generate particle IDs at the beginning of the simulation initialization
+template<class Dataset>
+void generateParticleIDs(Dataset& d, int rank, int numRanks)
+{
+    std::vector<size_t> ranksLocalParticles(numRanks);
+    size_t              localNumRanks = d.x.size();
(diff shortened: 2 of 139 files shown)


main/src/init/evrard_init.hpp
```diff
@@ -179,6 +179,8 @@ class EvrardGlassSphere : public ISimInitializer<Dataset>
 
         initEvrardFields(d, settings_);
 
+        generateParticleIDs(d, rank, numRanks);
+
```
sekelle: `generateParticleID` just initializes an additional field. It can be called in `initEvrardFields` where all the other fields are initialized as well. Same for the other test cases.

main/src/init/utils.hpp
```diff
@@ -120,6 +120,29 @@ void readFileAttributes(InitSettings& settings, const std::string& settingsFile,
     }
 }
 
+//! @brief generate particle IDs at the beginning of the simulation initialization
+template<class Dataset>
+void generateParticleIDs(Dataset& d, int rank, int numRanks)
```
sekelle: This function doesn't need the full dataset, only a `std::span<uint64_t> id` . At the call site `d.id` can be passed, which will either have size 0 if ids are (temporarily) disabled or have size `numLocalParticles`.

main/src/init/utils.hpp
```diff
@@ -120,6 +120,29 @@ void readFileAttributes(InitSettings& settings, const std::string& settingsFile,
...
+{
+    std::vector<size_t> ranksLocalParticles(numRanks);
```
sekelle: use `uint64_t` here to avoid compatibility issues with MacOS

main/src/init/utils.hpp
```diff
@@ -120,6 +120,29 @@ void readFileAttributes(InitSettings& settings, const std::string& settingsFile,
...
+    size_t              localNumRanks = d.x.size();
+    // fill ranksLocalParticles with the number of particles per rank
+    MPI_Allgather(&localNumRanks, 1, MPI_UNSIGNED_LONG, ranksLocalParticles.data(), 1, MPI_UNSIGNED_LONG,
```
sekelle: use `MpiType<uint64_t>{}` here instead of MPI_UNSIGNED_LONG

main/src/init/utils.hpp
```diff
@@ -120,6 +120,29 @@ void readFileAttributes(InitSettings& settings, const std::string& settingsFile,
...
+
+    for (int i = 0; i < rank; i++)
+    {
+        offset += ranksLocalParticles[i];
+    }
+
+    for (size_t i = 0; i < d.x.size(); i++)
+    {
+        d.id[i] = offset + i;
+    }
```
sekelle: These `for` loops are equivalent to:
```c++
std::exclusive_scan(ranksLocalParticles.begin(), ranksLocalParticles.end(), ranksLocalParticles.begin(), uint64_t(0));
std::iota(id.begin(), id.end(), ranksLocalParticles[rank]);
```
This will work also if `id` is not actually allocated, which is needed if we want to make it an optional feature later or temporarily disable it by removing `id` from the list of conserved fields.

sekelle (changes requested):
Please rebase on develop and squash into a single commit once done.

main/src/init/evrard_init.hpp
OsmanSeckinSimsek: this requires passing rank and numRanks to the init*Fields functions. Should I pass these parameters as well?

sekelle: You're already using the MPI communicator in `generateParticleIDs`. Might as well use it to obtain `rank` and `numRanks` as well. You can take the communicator as an argument to `generateParticleIDs` and pass `MPI_COMM_WORLD` at the call site. (One less function to remove `MPI_COMM_WORLD` from later).

sekelle:
Addressed in a follow-up PR

PR #453, opened 2024-10-08
add id field which is conserved to keep track of particles

OsmanSeckinSimsek:
It is branched from the latest develop and it includes a single commit, making it a cleaner PR.
I made the changes you suggested in the previous version, using exclusive_scan and iota, using gsl::span<uint64_t> id instead of passing the whole dataset, and moving the generateParticleIDs call inside the respective init*Fields for every testcase function.


sekelle (approved):
lgtm

PR #456, opened 2024-10-25
Fix box in cell property updates

sekelle:
Disallow calling LET cell property updates like counts, source centers and macs with a changed bounding box,
because changing the bounding box invalidates all node properties.


nknk567 (commented):
Successfully tested with the TDE simulation.

sekelle:
This is not quite ready yet. It works for the TDE simulation, but causes issues for the Evrard test.

sekelle:
After some further test, I think that Evrard is fine. In rare cases (e.g. `-n 500` on 32 ranks) it can happen that a few particles escape in the final expansion phase of the simulation. This can be avoided with slightly better gravity accuracy (`--theta 0.45` instead of the default `0.5`)

nknk567 (approved):

PR #457, opened 2024-10-29
Extend python IC generator with particle IDs

sekelle:


biddisco (approved):
LGTM, the initialConditions.h5 loads fine.

PR #477, opened 2025-01-16
ci

jgphpc:


sekelle (approved):
Let's get started with this :+1: 
We'll need to think how to check the outputs of the full application and we'll need 12 ranks or more for some tests.

PR #478, opened 2025-01-20
sound speed calculation; EOS with temperature and internal energy

nknk567:
Add factor of gamma when calculating sound speed (c = sqrt(gamma * (gamma - 1) * u))
Modify idealGas EOS such that it can be called with either temperature or internal energy u.
Make polytropic EOS more generally.
EOS Switch also in std hydro.

### DIFF
--- a/sph/include/sph/hydro_ve/eos_gpu.cu
+++ b/sph/include/sph/hydro_ve/eos_gpu.cu
@@ -42,9 +42,9 @@ namespace cuda
 {
 
 template<class Tt, class Tm, class Thydro>
-__global__ void cudaEOS(size_t firstParticle, size_t lastParticle, Tm mui, Tt gamma, const Tt* temp, const Tt* u,
-                        const Tm* m, const Thydro* kx, const Thydro* xm, const Thydro* gradh, Thydro* prho, Thydro* c,
-                        Thydro* rho, Thydro* p)
+__global__ void cudaComputeIdealGasEOS(size_t firstParticle, size_t lastParticle, Tm mui, Tt gamma, const Tt* temp,
+                                       const Tt* u, const Tm* m, const Thydro* kx, const Thydro* xm,
+                                       const Thydro* gradh, Thydro* prho, Thydro* c, Thydro* rho, Thydro* p)
 {
     unsigned i = firstParticle + blockDim.x * blockIdx.x + threadIdx.x;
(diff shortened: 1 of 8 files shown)


sph/include/sph/hydro_ve/eos_gpu.cu
```diff
@@ -59,22 +59,46 @@ __global__ void cudaEOS(size_t firstParticle, size_t lastParticle, Tm mui, Tt ga
...
+__global__ void cudaEOS_u(size_t firstParticle, size_t lastParticle, Tm mui, Tt gamma, const Tt* u, const Tm* m,
+                          const Thydro* kx, const Thydro* xm, const Thydro* gradh, Thydro* prho, Thydro* c, Thydro* rho,
+                          Thydro* p)
+{
+    unsigned i = firstParticle + blockDim.x * blockIdx.x + threadIdx.x;
+    if (i >= lastParticle) return;
+
+    Thydro p_i;
+    Thydro rho_i         = kx[i] * m[i] / xm[i];
+    util::tie(p_i, c[i]) = idealGasEOS_u(u[i], rho_i, gamma);
```
sekelle: I would put the if `u == nullptr` here and avoid two separate CUDA kernels. Performance-wise the difference is practically zero.

nknk567: Changed.

sph/include/sph/hydro_ve/eos_gpu.cu
```diff
@@ -111,5 +108,37 @@ void computeIsothermalEOS(size_t first, size_t last, Dataset& d)
...
+                                         Th* p, const Tmass* m, const Th* kx, const Th* xm, const Th* gradh, Th* prho,
+                                         Tt* temp, Th* c)
+{
+    unsigned i = first + blockDim.x * blockIdx.x + threadIdx.x;
+    if (i >= last) return;
+
+    Th rho_i        = kx[i] * m[i] / xm[i];
+    auto [p_i, c_i] = polytropicEOS(polytropic_const, polytropic_index, rho_i);
+    prho[i]         = p_i / (kx[i] * m[i] * m[i] * gradh[i]);
+    c[i]            = c[i];
```
sekelle, line 134: Should be `c[i] = c_i`

sekelle (approved):

PR #483, opened 2025-02-01
Particle subset selection

ChristopherBignamini:
@sekelle , this is the branch I'm using for the implementation of the particle subset selection. As you can see it is still a work in progress. Particularly, the following are the features/issue still open:

1) User definition of a spherical subdomain (one or more?) of interest (I'm working on this right now). For the moment I'm testing the code with a hardcoded id list in the main function.

2) The subdomain(s) must be stored in the form of attributes and should be generated with a Python script. 

2) Restart of a simulation using an existing checkpoint file: implementation of a procedure to handle the various cases, such as untagging of previously tagged particle and new subdomain(s) definition. 

3) Untagging of particles in the particle subset output file (the printed id have the most significant bit set to 1). 

4) GPU implementation of the particle tagging: I guess this could be optional since it is done only once.

5) Optimization of GPU/CPU implementations of selected particle position identification: here is where I used the stream compaction algorithm we discussed but it could be worth spend some time in finding possible performance issues. 

P.S.: the particle position search code (particle_selection.hpp and particle_selection.cu) is temporarily in the same folder of sphexa.cpp, I would need your help to identify the right place.


jgphpc:
cscs-ci run

jgphpc:
cscs-ci run

PR #484, opened 2025-02-05
enable GPU-resident data access for Ascent

jfavre:
If Ascent is compiled with CUDA support, it can directly access GPU-resident data

### DIFF
--- a/main/src/ascent_adaptor.h
+++ b/main/src/ascent_adaptor.h
@@ -112,13 +112,35 @@ void Execute(DataType& d, long startIndex, long endIndex)
     mesh["state/time"].set_external(&d.ttot);
 
     mesh["coordsets/coords/type"] = "explicit";
+#if defined(USE_CUDA)
+    mesh["coordsets/coords/values/x"].set_external(d.devData.x.data() + startIndex, endIndex - startIndex);
+    mesh["coordsets/coords/values/y"].set_external(d.devData.y.data() + startIndex, endIndex - startIndex);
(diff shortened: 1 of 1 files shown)


sekelle:
An issue is that "Density", `d.rho` or `d.devData.rho` is only populated on steps with file-output as the default `ve` propagator internally doesn't work with `rho`. Instead, density is `rho[i]  = kx[i] * m[i] / xm[i];`

An option could be to rename "Density" to "DensitySTD" and to introduce a composite field "Density" and define that as "kx * m / xm" if that is possible. Otherwise I would add "kx" and "xm" individually, then one might be able to plot `kx * m / xm` in the Ascent script.

jfavre:
> An issue is that "Density", `d.rho` or `d.devData.rho` is only populated on steps with file-output as the default `ve` propagator internally doesn't work with `rho`. Instead, density is `rho[i] = kx[i] * m[i] / xm[i];`
> 
> An option could be to rename "Density" to "DensitySTD" and to introduce a composite field "Density" and define that as "kx * m / xm" if that is possible. Otherwise I would add "kx" and "xm" individually, then one might be able to plot `kx * m / xm` in the Ascent script.

Thanks. Very good explanation. I think that for the moment, I would like to ask Ascent to do the math. This is a good exercise to see how much it costs. But presumably, we would do the operation on reduced data (clipped, or multi-res, or thresholded)

sekelle:
The data structure knows whether the GPU is active or not, so we don't need to duplicate the definition for the CPU and GPU cases. Also added `xm` and `kx`. Could you quickly test it?

sekelle (approved):

PR #485, opened 2025-02-08
Apply M2P for remote empty leaves

sekelle:
Remote empty leaves with non-zero multipole mass are guaranteed to pass MAC from
anywhere in the local domain. We may still get (rare) MAC failures when an empty part of a target particle group
bounding box protrudes outside the local domain. In this case the correct behavior is to pass MAC and apply M2P.

### DIFF
--- a/ryoanji/src/ryoanji/nbody/traversal_cpu.hpp
+++ b/ryoanji/src/ryoanji/nbody/traversal_cpu.hpp
@@ -96,14 +96,23 @@ void computeGravityGroup(const util::array<Vec4<T1>, N>& target, const TreeNodeI
      * the traversal routine to keep going. If the MAC passed, the multipole moments are applied
      * to the particles in the target box and traversal is stopped.
      */
-    auto descendOrM2P = [centers, multipoles, &target, &targetCenter, &targetSize, acc](TreeNodeIndex idx)
+    auto descendOrM2P =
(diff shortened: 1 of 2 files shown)


ryoanji/src/ryoanji/nbody/traversal_cpu.hpp
```diff
@@ -96,14 +96,23 @@ void computeGravityGroup(const util::array<Vec4<T1>, N>& target, const TreeNodeI
...
+    auto descendOrM2P =
+        [internalToLeaf, layout, centers, multipoles, &target, &targetCenter, &targetSize, acc](TreeNodeIndex idx)
     {
         const auto& com = centers[idx];
         const auto& mp  = multipoles[idx];
 
         bool violatesMac = cstone::evaluateMac(makeVec3(com), com[3], targetCenter, targetSize);
 
-        if (!violatesMac)
+        auto lidx = internalToLeaf[idx];
```
nknk567: lidx could be renamed to leafIdx so the same terminology is used in the CPU and GPU versions. Maybe also for isRemote.

nknk567 (changes requested):

nknk567 (approved):

ryoanji/src/ryoanji/nbody/traversal_cpu.hpp
sekelle: done

PR #488, opened 2025-02-14
Implement physics for protoplanetary disks / TDEs 

sekelle:


sekelle:
Continuation of https://github.com/sphexa-org/sphexa/pull/466

sekelle:
Hi Noah, I updated your PR with the latest develop and fixed some CMake issues. Please take these commits and recreate them with `git reset` and git commit` so they are yours again.

I think you don't need to fully duplicate the `std_disk` propagator. You could change the `std_hydro` prop to use `u` instead of temp, then inherit `std_disk` from `std_hydro`. The only function you need to override and reimplement is `integrate`.
(put `beta_cooling`, `du_timestep` and `central_force` there). `std_disk.hpp` can then be included and compiled in `std_hydro.cpp`

PR #494, opened 2025-03-31
Helmholtz EOS

OsmanSeckinSimsek:
adds Helmholtz EOS:
- CPU only version
- Uses helm_table.dat which includes all the tabulated values.
- add abar and zbar to particleData


rmcabezon (approved):

sph/include/sph/particles_data_gpu.cuh
```diff
@@ -200,6 +206,12 @@ public:
         whd = DevVector<HydroType>(whdTable.begin(), whdTable.end());
     }
 
+    void uploadHelmEOS(const sph::helmholtz_constants::Helmholtz_EOS* helmEOS)
+    {
+        cudaMalloc(&devHelmEOS, sizeof(helmEOS));
+        cudaMemcpy(devHelmEOS, helmEOS, sizeof(helmEOS), cudaMemcpyHostToDevice);
```
sekelle: Memory leak. In practice only leaks at the end of the simulation, but will still show up in memory sanitize tools.
Better to allocate this in the translation unit that implements the Helmoltz EOS and make an object to control the lifetime, and put that into the propagator, like we do for e.g. gravity/multipole_holder.

sekelle (changes requested):
* I'd recommend to put also the CPU version in a separate translation unit, not just the GPU version and thereby avoid the exposing all the global variables and symbols to the  rest of the code.

* Add stateful object that manages lifetime of data tables needed for the Helmoltz EOS, also in the translation unit.

* The helm_table is way too big to commit as source code, better to upload to Zenodo and grab from there whenever needed.

PR #495, opened 2025-04-11
make profiling on by default

OsmanSeckinSimsek:


sekelle (changes requested):
* Turning profiling on is a reasonable default when file output is enabled as well.
If there's no file output, I'd argue we can leave it safely off and avoid polluting the working directory when running quick tests.

* Energy measurements should not be on by default. These are not performance neutral. Can be done by changing the default value of `pmroot` to empty.

sekelle (approved):

PR #500, opened 2025-05-07
provides a much more up-to-date YAML example tuned for the wind-shock test

jfavre:
provides proper syntax for numerous options which can be tuned for other test cases, and more importantly,
eschew bug https://github.com/Alpine-DAV/ascent/issues/1494
Until this bug is fixed, *all* Ascent actions need to be compiled in, instead of read at runtime from the "ascent_actions.yaml" file.


OsmanSeckinSimsek (approved):

main/src/ascent_adaptor.h
```diff
@@ -126,16 +163,17 @@ void Execute(DataType& d, long startIndex, long endIndex)
...
+    addField(mesh, "alpha", get<"alpha">(d).data(), startIndex, endIndex);
+    addField(mesh, "m", get<"m">(d).data(), startIndex, endIndex);
+    //addField(mesh, "Smoothing Length", get<"h">(d).data(), startIndex, endIndex);
+    //addField(mesh, "Density", get<"rho">(d).data(), startIndex, endIndex);
+    //addField(mesh, "Internal Energy", get<"u">(d).data(), startIndex, endIndex);
+    //addField(mesh, "Pressure", get<"p">(d).data(), startIndex, endIndex);
+    //addField(mesh, "Speed of Sound", get<"c">(d).data(), startIndex, endIndex);
+    //addField(mesh, "ax", get<"ax">(d).data(), startIndex, endIndex);
+    //addField(mesh, "ay", get<"ay">(d).data(), startIndex, endIndex);
+    //addField(mesh, "az", get<"az">(d).data(), startIndex, endIndex);
```
sekelle, line 176: Is there a reason to comment these fields out or could we just leave them active?

sekelle (changes requested):
Minor question about commenting out some fields.

main/src/ascent_adaptor.h
jfavre: There had been times when Ascent would crash immediately and commenting out these would "fix" it (although I was not doing any use of these variables). I cannot reproduce a crash now, so, I guess we can activate these variables again.

sekelle (approved):

PR #502, opened 2025-05-30
Dfs stackless

sekelle:


domain/include/cstone/traversal/traversal.hpp
```diff
@@ -23,73 +23,49 @@
...
     {
-        for (int octant = 0; octant < 8; ++octant)
+        bool isLeaf  = childOffsets[node] == 0;
+        bool descend = not backtrack && continuationCriterion(node);
+        // process current node
+        if (isLeaf && descend) { endpointAction(node); }
+
+        TreeNodeIndex siblingIdx = (node - 1) % 8;
+        // determine next node
+        if (childOffsets[node] && descend) // can we descend?
```
nknk567: Could be if (!isLeaf && descend), then it's clearer with the condition before.

sekelle: done

PR #504, opened 2025-06-23
Fix atomicMinFloat, add atomicMaxFloat and tests for both

fthaler:
`atomicMinFloat` previously failed in corner cases involving values of `-0.0f`. This adds tests for these special cases and also the missing `atomicMaxFloat`. Based on https://stackoverflow.com/a/72461459


sekelle (approved):
lgtm

PR #507, opened 2025-06-24
Fix acceleration timestep

nknk567:
This changes the way the acceleration timestep is computed by using the smoothing length which is also used as the softening in gravity instead of the constant eps. dt = etaAcc * sqrt(h / a)


sekelle:
acc-ts-fix

Slight differences to previous version, but seems sane.

sekelle (approved):

PR #510, opened 2025-07-05
improved IC generator overwrites existing data 

lks1248:
The commit dd4b678 optimizing the IC generator changed the behaviour of `assembleCuboid` such that existing data in the position vectors is overridden. 

For Kelvin-Helmholtz this is not an issue and is easily fixed, however future tests (e.g the triple point shock) would benefit from being able to sequentially assemble cuboids. I'm opening a PR as I don't know what the best resolution for this is.


sekelle (approved):
Original cumulative behavior restored.

PR #516, opened 2025-07-29
Fixed Boundary Conditions

lks1248:
Adds fixed boundary conditions for the rim of cuboid boxes. A sigmoid function is used to remove acceleration of particles within `8*h` of the boundary.

Is a prerequisite of the Rayleigh-Taylor, Sod shock and Triple-Point shock test cases.

### DIFF
--- a/sph/include/sph/positions.hpp
+++ b/sph/include/sph/positions.hpp
@@ -53,7 +53,7 @@ HOST_DEVICE_FUN Tc fbcAdjustFactor(const cstone::Vec3<Tc> X, const cstone::Box<T
         constexpr Th width    = 2.;
         constexpr Th invWidth = 1 / width;
         // width of particles that are held at zero, in smoothing lengths
-        constexpr Th       offset          = 2.;
+        constexpr Th       offset          = 1.;
         cstone::Vec3<bool> isBoundaryFixed = {
             box.boundaryX() == cstone::BoundaryType::fixed,
             box.boundaryY() == cstone::BoundaryType::fixed,
--- a/sph/include/sph/positions.hpp
+++ b/sph/include/sph/positions.hpp
@@ -144,10 +144,10 @@ void updateTempHost(size_t startIndex, size_t endIndex, Dataset& d, const cstone
 
     bool haveMui      = !d.mui.empty();
     auto constCv      = idealGasCv(d.muiConst, d.gamma);
-    auto adjustForFBC = 1.0;
+    auto adjustForFBC = T(1);
 
 #pragma omp parallel for schedule(static)
-    for (size_t i = startIndex; i < endIndex; i++)
+    for (std::size_t i = startIndex; i < endIndex; i++)
     {
         if (anyFBC) { adjustForFBC = min(fbcAdjustFactors({d.x[i], d.y[i], d.z[i]}, box, d.h[i])); }
         auto cv    = haveMui ? idealGasCv(d.mui[i], d.gamma) : constCv;
@@ -155,7 +155,7 @@ void updateTempHost(size_t startIndex, size_t endIndex, Dataset& d, const cstone
         // notice the common factor of dt in energyUpdate: to apply the Fixed Boundary Correction we can do it on dt.
         // we divide out dt_m1 by that factor so it applies only once to each of the updating terms
         d.temp[i] =
-            energyUpdate(u_old, d.minDt * adjustForFBC, d.minDt_m1 * (1. / adjustForFBC), d.du[i], d.du_m1[i]) / cv;
+            energyUpdate(u_old, d.minDt * adjustForFBC, d.minDt_m1 * (T(1) / adjustForFBC), d.du[i], d.du_m1[i]) / cv;
         d.du_m1[i] = d.du[i];
     }
 }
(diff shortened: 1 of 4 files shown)


sph/include/sph/positions.hpp
```diff
@@ -42,12 +42,38 @@
...
+
+            Th relDistanceMax = std::abs(boxMax[j] - X[j]) / hi;
+            Th relDistanceMin = std::abs(boxMin[j] - X[j]) / hi;
+            Th minDistance    = relDistanceMin < relDistanceMax ? relDistanceMin : relDistanceMax;
+
+            if (minDistance < 2 * threshold)
+            {
+                Tc correction = 0.5 * (std::tanh(4 * minDistance * invTHold - 4) + 1);
+                A[j] *= correction;
+                V_nm[j] *= correction;
```
lks1248: Correcting velocities is not strictly required if you set up initial velocities within the same `4*h` as zero as well. It will be required if you have solid boundaries that evolve with time, for example.

sph/include/sph/positions.hpp
```diff
@@ -42,21 +42,90 @@
 
 namespace sph
 {
+//! @brief checks whether a particle is close to a fixed boundary and reduces a scalar if so
+template<class Tc, class Th, class Ts>
+HOST_DEVICE_FUN void fbcAdjustScalar(const cstone::Vec3<Tc> X, Ts& scalar, const cstone::Box<Tc>& box, const Th hi)
```
sekelle: It's generally much cleaner in a case like this to take `scalar` by value and return the modified result and assign that at the call site. Otherwise you don't see as clearly what's going on at the call site, i.e. you have to go look into this function to see you're modifying a parameter by reference.

sph/include/sph/positions.hpp
```diff
@@ -42,21 +42,90 @@
...
+        }
+    }
+}
 
-//! @brief checks whether a particle is in the fixed boundary region in one dimension
+//! @brief checks whether a particle is close to a fixed boundary and reduces the acceleration if so
 template<class Tc, class Th>
-HOST_DEVICE_FUN bool fbcCheck(Tc coord, Th h, Tc top, Tc bottom, bool fbc)
+HOST_DEVICE_FUN void fbcAdjustVector(const cstone::Vec3<Tc> X, cstone::Vec3<Tc>& V_nm, cstone::Vec3<Tc>& A,
+                                     const cstone::Box<Tc>& box, const Th hi)
```
sekelle: This is a a) duplicate of the scalar version, which could be re-used in here by calling component by component and b) won't be needed if you compute the correction factor once for the particle and apply the factor where needed.

sph/include/sph/positions.hpp
```diff
@@ -42,21 +42,90 @@
...
+    // width of particles that are held at zero, in smoothing lengths
+    constexpr Th       offset          = 2.;
+    cstone::Vec3<bool> isBoundaryFixed = {
+        box.boundaryX() == cstone::BoundaryType::fixed,
+        box.boundaryY() == cstone::BoundaryType::fixed,
+        box.boundaryZ() == cstone::BoundaryType::fixed,
+    };
+    cstone::Vec3<Tc> boxMax = {box.xmax(), box.ymax(), box.zmax()};
+    cstone::Vec3<Tc> boxMin = {box.xmin(), box.ymin(), box.zmin()};
+
```
sekelle, line 45: `Vec3` has vectorized min/max/abs functions:

```suggestion
cstone::Vec3<Tc> minDistance = min(abs(boxMax - X), abs(boxMin - X)) * (Th(1.0) / hi);
```

sph/include/sph/positions.hpp
sekelle: This function should have a unit-test.

sph/include/sph/positions.hpp
```diff
@@ -42,21 +42,90 @@
...
+    for (int j = 0; j < 3; ++j)
+    {
+        if (isBoundaryFixed[j])
```
sekelle:
```suggestion
        if (isBoundaryFixed[j] && minDistance[j] < 2 * width + offset)
```

sph/include/sph/positions.hpp
sekelle: The correction of a particle only depends on the box, the particle position and smoothing length.
You should separate the computation of the factor from applying it to energies, velocities, etc.

* Calculate correction factor for each particle
* Apply original energy, velocity updates, rather than putting all the extra arguments into their signatures
* Multiply by correction factor where needed.

sekelle (changes requested):

sekelle (changes requested):
Fixed-boundary correction factors per particle can be separated from applying them.

sekelle (changes requested):
Fixed-boundary correction factors per particle can be separated from applying them.

sekelle (changes requested):
Fixed-boundary correction factors per particle can be separated from applying them.

sekelle (changes requested):
Fixed-boundary correction factors per particle can be separated from applying them.

sekelle (changes requested):
Fixed-boundary correction factors per particle can be separated from applying them.

sekelle (approved):

PR #530, opened 2025-08-21
Bencher support

jgphpc:


sekelle:
Can be moved to cscs-reframe-tests.

PR #532, opened 2025-08-25
Fix du timestep, introduce additional timestep, Einstein potential

nknk567:
Fix error in du timestep, introduce an additional timestep to resolve orbit around central object in thin gas, introduce Einstein potential


physics/Disk/include/central_force_loop.hpp
sekelle, line 1: Perhaps `central_potential.hpp` would be a better name?

sekelle (approved):

PR #533, opened 2025-09-08
Avoid duplication of particle data fields on the CPU and GPU

sekelle:
Useful for unified memory architectures such as the MI300A.

### DIFF
--- a/main/src/propagator/ipropagator.hpp
+++ b/main/src/propagator/ipropagator.hpp
@@ -170,6 +170,26 @@ protected:
         output(first, last, simData.chem, writer);
     }
 
+    void logDomainStats(const DomainType& domain, ParticleDataType& simData)
+    {
+        timer.logStatistics("numParticles", domain.nParticles());
+        timer.logStatistics("numHalos", domain.nParticlesWithHalos() - domain.nParticles());
+        timer.logStatistics("assignment", domain.assignmentStart());
+
+        auto hostMem = simData.hydro.memStats();
+        timer.logStatistics("hostMemSizeBytes", hostMem[1]);
+        timer.logStatistics("hostCapSizeBytes", hostMem[2]);
+
+        using AccType = ParticleDataType::AcceleratorType;
+        if constexpr (cstone::HaveGpu<AccType>{})
+        {
+            auto devMem = simData.hydro.devData.memStats();
+            timer.logStatistics("devMemSizeBytes", devMem[1]);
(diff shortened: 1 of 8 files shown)


main/src/propagator/ipropagator.hpp
```diff
@@ -168,6 +169,26 @@ class Propagator
         output(first, last, simData.chem, writer);
     }
 
+    void logDomainStats(const DomainType& domain, ParticleDataType& simData)
```
ChristopherBignamini, line 172: Since only the hydro (ParticlesData<AccType>)  field of the ParticleDataType is used inside the function, does it make sense to pass only that field instead of the entire type? I see it's passed by reference, my comment is more about having a minimal signature (plus I'd like to understand which approach in general you prefer: minimize the signatures or minimize dependencies wrt externally defined types).

sph/include/sph/particles_data_gpu.cuh
```diff
@@ -184,6 +184,23 @@ public:
...
+                std::visit([]<class V>(V* arg) { return sizeof(typename V::value_type) * arg->size(); }, data_[i]);
+            sumOfCap +=
+                std::visit([]<class V>(V* arg) { return sizeof(typename V::value_type) * arg->capacity(); }, data_[i]);
+        }
+
+        std::size_t free{0}, total{0};
+        cudaMemGetInfo(&free, &total);
+        return {size(), sumOfSize, sumOfCap, free, total};
+    }
+
```
ChristopherBignamini, line 203: Does it make sense/Is it possible to do a bit of refactoring of this function with the corresponding one of particles_data.hpp? If I'm not wrong their implementation is similar except for the call to cudaMemGetInfo.

sekelle: Do you mean reduce code duplication?

main/src/propagator/ipropagator.hpp
sekelle: For free functions I think it's always desirable to minimize the signatures. In this case, the function is a class member, and `DomainType` a class template, so you could argue either way. Because we might add statistics for more than just hydro, I decided to use the class template.

sph/include/sph/particles_data_gpu.cuh
ChristopherBignamini: Yes, I worded that badly, my apologies.

ChristopherBignamini (approved):
lgtm

PR #534, opened 2025-09-11
TDE initializers

nknk567:
Contains TDE initializer and polytrope initializer.
Also the file split initializer is updated such that it handles the temp or u fields.


main/src/init/file_init.hpp
```diff
@@ -225,14 +225,18 @@ class FileSplitInit : public ISimInitializer<Dataset>
...
+        if (d.isAllocated("temp")) { replicateField(reader, "temp", d.temp, T(1)); }
+        else if (d.isAllocated("u")) { replicateField(reader, "u", d.u, T(1)); }
 
         std::fill(d.du_m1.begin(), d.du_m1.end(), 0);
         std::fill(d.rung.begin(), d.rung.end(), 0);
         std::transform(d.vx.begin(), d.vx.end(), d.x_m1.begin(), [dt = d.minDt](auto v_) { return v_ * dt; });
         std::transform(d.vy.begin(), d.vy.end(), d.y_m1.begin(), [dt = d.minDt](auto v_) { return v_ * dt; });
         std::transform(d.vz.begin(), d.vz.end(), d.z_m1.begin(), [dt = d.minDt](auto v_) { return v_ * dt; });
 
+        generateParticleIDs(d.id);
```
sekelle: This will overwrite all particle IDs upon restart from file. These IDs have to be preserved across restarts.

main/src/init/file_init.hpp
```diff
@@ -225,14 +225,18 @@ class FileSplitInit : public ISimInitializer<Dataset>
         replicateField(reader, "vx", d.vx, T(1));
         replicateField(reader, "vy", d.vy, T(1));
         replicateField(reader, "vz", d.vz, T(1));
-        replicateField(reader, "temp", d.temp, T(1));
+
+        if (d.isAllocated("temp")) { replicateField(reader, "temp", d.temp, T(1)); }
+        else if (d.isAllocated("u")) { replicateField(reader, "u", d.u, T(1)); }
```
sekelle, line 248: This will overwrite temperatures or internal energies for all simulations upon restart, which will cause them to break.

sekelle (changes requested):

main/src/init/file_init.hpp
sekelle: Misread, it's fine.

main/src/util/timer.hpp
```diff
@@ -135,7 +135,7 @@ class Timer
     using SupportedTypes   = util::TypeList<float, double, uint32_t, uint64_t, int32_t, int64_t>;
     using SupportedVariant = util::Reduce<std::variant, util::Map<std::vector, SupportedTypes>>;
 
-    std::map<std::string, SupportedVariant> perfStats;
+    std::map<std::string, SuppitortedVariant> perfStats;
```
sekelle: typo

main/src/propagator/std_hydro.hpp
```diff
@@ -178,6 +178,8 @@ class HydroProp : public Propagator<DomainType, DataType>
         size_t first = domain.startIndex();
         size_t last  = domain.endIndex();
 
+        relaxSystem(first, last, d);
```
sekelle: Doesn't this break all the other test cases where `d.relaxationTimescale = 0` by default?

nknk567: It's guarded, in relaxSystem() there is: if (d.relaxationTimescale <= 0.) return;

We could also remove it from the propagator by default and just have relaxSystem() defined so it can be used if needed.

main/src/util/timer.hpp
nknk567: fixed

main/src/init/polytrope_init.hpp
```diff
@@ -0,0 +1,206 @@
...
+        Dataset d;
+        settings_ = buildSettings(d, polytropeConstants(), settingsFile, reader);
+
+        const double r                   = settings_["polytrope::r"];
+        const double mTotal              = settings_["polytrope::mTotal"];
+        const double gravConstant        = settings_["gravConstant"];
+        const double t_relax             = std::sqrt(r * r * r / (gravConstant * mTotal)) / 3.;
+        settings_["relaxationTimescale"] = t_relax;
+
+        readSingleAttribute("relaxationTimescale", settingsFile, reader);
```
sekelle: What are you trying to do here, why do you need to read this setting again?

sekelle (approved):
Can be merged once CI passes.

PR #537, opened 2025-09-29
Id tagging and finding basic functionalities 

ChristopherBignamini:
This PR contains the basic id tagging and finding components of the particles_subset_gpu branch (https://github.com/sphexa-org/sphexa/pull/483) with the corresponding unit tests. At this stage the id tagging/finding feature is not part of a SPH-EXA run: that part will be included in a following PR. 

@sekelle, as you can see there are two implementations for tagged ids finding in the .cpu e .cu files: I temporarily left in the code two old implementations because they seem to be faster, at least on a "synthetic" set of ids. I can just keep the one you proposed for the CPU implementation and is GPU counterpart, implementing the same strategy, and work on the performances in a later stage.

### DIFF
--- a/main/src/io/id_tag_utils.cpp
+++ b/main/src/io/id_tag_utils.cpp
@@ -40,24 +40,19 @@
 namespace sphexa
 {
 
-void applyTaggingMask(IdType groupId, IdType& id)
+void applyTaggingMask(uint64_t groupId, uint64_t& id)
 {
-    if (groupId >= supGroupId)
-        throw std::runtime_error("Tagging group id larger than max value (" + std::to_string(supGroupId) + ")\n");
+    if (groupId >= maxNumGroupIds)
+        throw std::runtime_error("Tagging group id larger than max value (" + std::to_string(maxNumGroupIds) + ")\n");
 
-    // Clear previous tagging if present
-    if(IsMasked{}(id)) {
-        id &= ~taggingCheckMask;
-        std::cout << "Warning: applying tagging mask to already tagged id " << id << "\n";
-    }
+    // Clear previous tagging bits if any
+    id &= ~taggingCheckMask;
 
-    // Apply new tagging information
     id |= ((groupId+1) << taggingMaskStartingBit);
 }
 
-template<class IdTypeP>
-void tagIdsInList(std::span<IdTypeP> ids, size_t first, size_t last, std::span<const IdTypeP> selectedIds,
-    const LocalIndex groupId)
+void tagIdsInList(std::span<uint64_t> ids, size_t first, size_t last, std::span<const uint64_t> selectedIds,
+    const uint32_t groupId)
 {
     const auto idListBeginIt = ids.begin()+first;
     const auto idListEndIt = ids.begin()+last;
@@ -70,14 +65,10 @@ void tagIdsInList(std::span<IdTypeP> ids, size_t first, size_t last, std::span<c
         }
     });
 }
-template void tagIdsInList<IdType>(std::span<IdType> ids, size_t first, size_t last, std::span<const IdType> selectedIds,
-    const LocalIndex groupId);
 
-
-template<class IdTypeP>
-void tagIdsInSphere(std::span<IdTypeP> ids, std::span<const CoordinateType> x, std::span<const CoordinateType> y,
+void tagIdsInSphere(std::span<uint64_t> ids, std::span<const CoordinateType> x, std::span<const CoordinateType> y,
     std::span<const CoordinateType> z, size_t firstIndex, size_t lastIndex, IdSelectionSphere selSphereData,
-    const LocalIndex groupId)
+    const uint32_t groupId)
 {
     const auto squareRadius = selSphereData[3]*selSphereData[3];
     const auto sphereCenter = util::makeVec3(selSphereData);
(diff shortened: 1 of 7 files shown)


main/src/io/id_tag_utils.hpp
```diff
@@ -0,0 +1,144 @@
...
+        return (id & taggingCheckMask) != 0;
+    }
+};
+
+/*! @brief Application of tagging mask to a given id
+ *
+ * @param[in]  groupId      selection group id (must be < supGroupId)
+ * @param[out] id           input id with tagging mask applied
+ */
+void applyTaggingMask(IdType groupId, IdType& id);
```
sekelle: Avoid modifying references when you can. Better to take the id by value and return the modified tag.

main/src/io/id_tag_utils.hpp
```diff
@@ -0,0 +1,144 @@
...
+#include <vector>
+
+#include "cstone/tree/definitions.h"
+#include "sph/types.hpp"
+
+namespace sphexa
+{
+
+using IdType = uint64_t;
+using LocalIndex = uint32_t;
```
sekelle: Should not be needed. Just keep `LocalIndexP` as a template and instantiate the templates with `uint32_t` and uint64_t`.

main/src/io/id_tag_utils.hpp
```diff
@@ -0,0 +1,144 @@
...
+using CoordinateType = sph::SphTypes::CoordinateType;
+
+/*! @brief Number of bits used for tagging information storage
+ */
+constexpr IdType taggingMaskSize = 10;
```
sekelle: Suggest rename: `tagNumBits`

main/src/io/id_tag_utils.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+/*! @brief Given the taggingMaskSize, this is the maximum selection group id value plus one
+ */
+constexpr IdType supGroupId = (1 << taggingMaskSize) - 1;
```
sekelle: Suggest rename: `maxNumGroupIds`

main/src/io/id_tag_utils.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+using IdType = uint64_t;
+using LocalIndex = uint32_t;
+using CoordinateType = sph::SphTypes::CoordinateType;
+
+/*! @brief Number of bits used for tagging information storage
+ */
+constexpr IdType taggingMaskSize = 10;
+
+/*! @brief Given the taggingMaskSize, this is the maximum selection group id value plus one
```
sekelle: "Given tagNumBits, the maximum number of groups we can address is 2^tagNumBits - 1. We subtract one, because groupId=0 corresponds to an unmasked particle ID."

main/src/io/id_tag_utils.hpp
```diff
@@ -0,0 +1,144 @@
...
+/*! @brief Tagged id (in first:last range) identification, GPU version
+ *
+ * @param[in]  ids          ordered id list
+ * @param[in]  first        first id index // TODO number of elements and pass iterator?
+ * @param[in]  last         last (excluded) id index
+ * @param[out] taggedIdsIndexes  vector of indexes of tagged ids
+ */
+template<class IdTypeP, class LocalIndexP>
+void findTaggedIdsGPU(std::span<const IdTypeP> ids, size_t first, size_t last, std::vector<LocalIndexP>& taggedIdsIndexes);
+extern template void findTaggedIdsGPU<IdType, LocalIndex>(std::span<const IdType> ids, size_t first, size_t last, std::vector<LocalIndex>& taggedIdsIndexes);
```
sekelle: Misunderstanding: the template instantiation belongs into .cu or .cpp file. Here you only need:
```c++
template<class IdTypeP, class LocalIndexP>
extern void findTaggedIdsGPU(std::span<const IdTypeP> ids, size_t first, size_t last, std::vector<LocalIndexP>& taggedIdsIndexes);
```

main/src/io/id_tag_utils.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+
+/*! @brief Id tagging spherical volume definition
+ * (center[0:2], radius)
+ */
+using IdSelectionSphere = cstone::Vec4<CoordinateType>;
+
+/*! @brief Id tagging list definition
+ */
+using IdSelectionList = std::vector<IdType>;
```
sekelle: Seems not used?

main/src/io/id_tag_utils.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+void applyTaggingMask(IdType groupId, IdType& id)
+{
+    if (groupId >= supGroupId)
+        throw std::runtime_error("Tagging group id larger than max value (" + std::to_string(supGroupId) + ")\n");
+
+    // Clear previous tagging if present
+    if(IsMasked{}(id)) {
+        id &= ~taggingCheckMask;
+        std::cout << "Warning: applying tagging mask to already tagged id " << id << "\n";
```
sekelle: I would remove the old mask unconditionally without warning. Periodically selecting new spheres during the simulations will most likely be the default use cause.

sekelle (changes requested):
Needs clang-format as well.

main/src/io/id_tag_utils.hpp
ChristopherBignamini: I'm temporarily keeping this definition to simplify the interface of tagIdsInList with multiple lists, I will delete it once we define the final signature.

main/src/io/id_tag_utils.cpp
```diff
@@ -0,0 +1,120 @@
...
+            }
+        });
+        groupId++;
+    }
+}
+
+void tagIdsInSphere(std::span<uint64_t> ids, std::span<const CoordinateType> x, std::span<const CoordinateType> y,
+    std::span<const CoordinateType> z, size_t firstIndex, size_t lastIndex, std::span<const IdSelectionSphere> selSphereData)
+{
+    uint64_t groupId = 0; // TODO: how do we decide the starting groupId?
```
ChristopherBignamini: @sekelle which strategy do you think we should adopt for groupId definition in case the tagIdsInSphere/List function is called multiple times during a simulation?

sekelle: I think we should change the signature to
```c++
void tagIdsInSphere(std::span<uint64_t> ids, std::span<const CoordinateType> x, std::span<const CoordinateType> y,
    std::span<const CoordinateType> z, size_t firstIndex, size_t lastIndex, std::span<const IdSelectionSphere> selSphereData, std::span<unsigned> sphereGroupIds)
```

And the same for `tagIdsInList`; instead of the span of vectors, just take a single span for all particles to be tagged and a second span with the groupId for each particle to be tagged.

sekelle:
`findTaggedIdsGPU` should return the extracted Ids in a GPU `DeviceVector`, because in order to extract the particle data from the fields on GPUs, we will run `gather` on the device, providing the output from `findTaggedIdsGpu` as the argument. At the call site, `cstone::DeviceVector` can be used for that, the implementation can use a template `DeviceVector` in the signature and an instatiation in the `.cu` file with `cstone::DeviceVector`.

sekelle (approved):

PR #538, opened 2025-09-29
Fix misplaced view paths in spack

jgphpc:


ci/cuda12/run_debug_sm90.yml
```diff
@@ -3,8 +3,6 @@ include:
 
 variables:
   MYUENV: "prgenv-gnu/25.06:rc5"
-  PATH1: "sphexa+spack/view/bin"
-  PATH2: "sphexa+spack/view/sbin"
```
sekelle, line 7: Deserves a comment that (and why?) the spack views don't work and why these variables can't be set for each test right here.

I would also suggest to change names:

`PATH2` -> `TEST_INSTALL_DIR`
`PATH1` -> `APP_INSTALL_DIR`

sekelle (changes requested):

sekelle (approved):

sekelle:
Config looks ok to me, but CI still fails due to missing executables.

PR #542, opened 2025-10-14
New Neighborhood API

fthaler:
Introduces a new neighbor lists API into the corner stone library and enables its use in sphexa.

Summary of all changes:
- Introduces `cstone::ijloop` namespace which includes a new API for particle-particle neighbor search.
- Introduces the concept of neighborhoods which provide a common API for particle-particle neighbor search on CPU and GPU.
- Uses the new API for the existing full CPU neighbor list, now available as `cstone::ijloop::CpuFullNeighborListNeighborhood`.
- Uses the new API for the existing GPU neighbor list, `cstone::ijloop::GpuAlwaysTraverseNeighborhood` which wraps the existing fast warp-based tree traversal in the new API.
- Adapts the `std`, `ve`, and `ve-bdt` propagators to use the new neighborhoods.
- Introduces unit tests for the new neighborhoods.
- Note: all particle-particle distances are now computed in double precision, while previously they were computed in double precision inside the tree traversal code but re-computed in single precision in some kernels. This can lead to noticeable numerical differences in sensitive kernels (obviously with higher precision in the new implementation).


sekelle (approved):

domain/include/cstone/util/uninitialized.hpp
sekelle: Will this be used in the clustered implementation or is it no longer needed?

PR #547, opened 2025-11-13
Fix KH growth-rate observable check

sekelle:


iomaganaris (approved):
LGTM 👍 Thanks for the quick fix

PR #549, opened 2025-11-18
Add rocthrust and hipcub dependencies in CMake to fix HIP spack package build

sekelle:


main/src/observables/conserved_gpu.cu
```diff
@@ -44,6 +43,25 @@ namespace sphexa
...
+{
+    template<std::size_t... Is>
+    HOST_DEVICE_FUN Tuple operator()(const Tuple& a, const Tuple& b, std::index_sequence<Is...>)
+    {
+        return Tuple((get<Is>(a) + get<Is>(b))...);
+    }
+};
+
+template<class Tuple>
+struct TuplePlus
```
fthaler, line 46: Maybe just implement this for the specific use case here (`thrust::tuple<double, Vec3<double>, Vec3<double>>`)? Saves the `TuplePlusImpl`.

main/src/observables/gpu_reductions.cu
```diff
@@ -33,12 +33,30 @@
...
+{
+    template<std::size_t... Is>
+    HOST_DEVICE_FUN Tuple operator()(const Tuple& a, const Tuple& b, std::index_sequence<Is...>)
+    {
+        return Tuple((get<Is>(a) + get<Is>(b))...);
+    }
+};
+
+template<class Tuple>
+struct TuplePlus
```
fthaler, line 45: Same as above, only implement it for `thrust::tuple<double, double, double>`.

fthaler (changes requested):
Two minor possible simplifications spotted, rest looks good.

main/src/observables/gpu_reductions.cu
sekelle: done

fthaler (approved):

PR #550, opened 2025-11-18
Update CMake CUDA architectures in README

rmcabezon:
Removed architecture '85' from CMake configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,9 @@ The code requires a **C++20 compiler** for both the CPU and GPU parts.
 * CUDA 12 and later
 * ROCm 6 and later. ROCm 5 compiles, but has bugs preventing the reliable use of GPU-aware-MPI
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -85,7 +85,7 @@ ml CUDA/12.8.0
 
 mkdir build
 cd build
-cmake -DCMAKE_CUDA_ARCHITECTURES="80;85;90" <GIT_SOURCE_DIR>
```
sekelle, line 88: We could add a line to the readme to check the architecture of your GPU here: https://developer.nvidia.com/cuda-gpus

sekelle (changes requested):

README.md
rmcabezon: Sure! Good point. I'll add it.

sekelle (approved):

PR #551, opened 2025-11-21
Separate halo and MAC peer rank lists

sekelle:


domain/include/cstone/focus/peer_flags.hpp
```diff
@@ -0,0 +1,94 @@
...
+{
+    auto globalTreeActive = globalTree;
+    if constexpr (useGpu)
+    {
+        std::vector<KeyType> globalTreeBackingBuffer;
+        globalTreeBackingBuffer.resize(globalTree.size());
+        memcpyD2H(globalTree.data(), globalTree.size(), globalTreeBackingBuffer.data());
+        globalTreeActive = std::span(globalTreeBackingBuffer);
+    }
+    return focusPeers<KeyType>(globalOffsets, focusOffsets, myRank, globalTreeActive, focusTree);
```
fthaler: Use-after-free bug when `useGpu` is true: `globalTreeBackingBuffer` is destructed on closing block scope on line 81, but its buffer is reused on line 82.

domain/include/cstone/halos/halo_peers.hpp
```diff
@@ -0,0 +1,50 @@
...
+    int numRanks = fAssignment.size();
+    std::vector<int> peerFlags(numRanks, 0);
+#pragma omp parallel for
+    for (int rank = 0; rank < numRanks; ++rank)
+    {
+        if (rank == myRank) { continue; }
+
+        TreeNodeIndex focStart = fAssignment[rank].start();
+        TreeNodeIndex focEnd   = fAssignment[rank].end();
+        if (focEnd < focStart) { focEnd = focStart; }
```
fthaler:
```suggestion
        if (focEnd < focStart) { continue; }
```

domain/include/cstone/halos/halos.hpp
```diff
@@ -109,6 +109,8 @@ class Halos
     RecvList incomingHaloIndices_;
     SendList outgoingHaloIndices_;
 
+    std::vector<int> extPeers_, intPeers_;
```
fthaler: I’m not a big fan of these names as it’s not intuitive to me which is which (does the an external peer require data from us or does it own some external data that we need locally?). Further, the abbreviations ‘ext’ and especially ‘int’ are ambiguous. Maybe `dstPeers` and `srcPeers` or something similar which clarifies the transfer direction?

domain/test/integration_mpi/exchange_general_gpu.cpp
```diff
@@ -101,7 +101,7 @@ static void generalExchangeRandomGaussian(int thisRank, int numRanks)
 
     // *******************************
 
-    auto peers = findPeersMac(thisRank, assignment, domainTree.cdata(), box, invThetaEff);
+    // auto peers = findPeersMac(thisRank, assignment, domainTree, box, invThetaEff);
```
fthaler:
```suggestion
```

fthaler (changes requested):

domain/include/cstone/focus/peer_flags.hpp
sekelle: Right, good spot.

domain/include/cstone/halos/halo_peers.hpp
sekelle: The line can be deleted.

domain/include/cstone/halos/halos.hpp
sekelle: I agree that there might be better names. Int/Ext stands for interior/exterior. Src/dest or send/recv is not a good name here, because messages are exchanged in either direction, i.e.

exterior -> interior: treelet exchange ("let remote ranks now our LET structure for the part that covers their subdomain")
interior -> exterior: cell properties ("receive cell properties for the non-local areas in our LET")

Exterior ranks = the ranks that own non-local cells in our LET which do not appear in the global tree
Interior ranks = the ranks that have non-local cells in their LET that fall into our subdomain

fthaler: Makes sense. How about adding the definition of exterior peers and interior peers in a comment in the code? And maybe also using `exteriorPeers` and `interiorPeers` to avoid ambiguity?

sekelle: Done

fthaler (approved):

PR #552, opened 2025-11-24
Custom MPI reduction

sekelle:
Prevents overflow issues that may occur when the global octree is not close to converged during first build.

### DIFF
--- a/domain/include/cstone/domain/assignment.hpp
+++ b/domain/include/cstone/domain/assignment.hpp
@@ -69,9 +69,6 @@ public:
             d_csTree_     = leaves_;
             d_nodeCounts_ = nodeCounts_;
             buildOctreeGpu(d_csTree_.data(), tree_.data());
-
-            hostTree_.resize(nNodes(leaves_));
-            updateInternalTree<KeyType>(leaves_, hostTree_.data());
         }
         else { updateInternalTree<KeyType>(leaves_, tree_.data()); }
     }
@@ -125,10 +122,10 @@ public:
 
         if constexpr (gpu)
         {
-            hostTree_.resize(tree_.numLeafNodes);
-            memcpyD2H(tree_.prefixes.data(), tree_.prefixes.size(), hostTree_.prefixes.data());
-            memcpyD2H(tree_.childOffsets.data(), tree_.childOffsets.size(), hostTree_.childOffsets.data());
-            std::copy_n(tree_.levelRange.data(), tree_.levelRange.size(), hostTree_.levelRange.data());
+            reallocate(leaves_, d_csTree_.size(), growthRate_);
+            reallocate(nodeCounts_, d_nodeCounts_.size(), growthRate_);
+            memcpyD2H(d_csTree_.data(), d_csTree_.size(), leaves_.data());
+            memcpyD2H(d_nodeCounts_.data(), d_nodeCounts_.size(), nodeCounts_.data());
         }
 
         auto newAssignment = makeSfcAssignment(numRanks_, nodeCounts_, leaves_.data());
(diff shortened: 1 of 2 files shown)


domain/include/cstone/tree/update_mpi_gpu.cuh
```diff
@@ -28,24 +28,34 @@
...
 
+inline void summax(void* inP, void* inoutP, int* len, MPI_Datatype*)
+{
+    auto* in    = reinterpret_cast<unsigned*>(inP);
+    auto* inout = reinterpret_cast<unsigned*>(inoutP);
+    for (int i = 0; i < *len; ++i)
+    {
+        auto a   = in[i];
+        auto b   = inout[i];
+        inout[i] = std::max(a + b, std::max(a, b));
```
fthaler:
```suggestion
        if (__builtin_uadd_overflow(a, b, &inout[i])) {
            inout[i] = std::numeric_limits<unsigned>::max();
        }
```
There is a compiler builtin (supported by Clang and GCC) which could be used here. Further, might using `std::numeric_limits<unsigned>::max()` instead of `std::max(a, b)` improve convergence?

domain/include/cstone/domain/assignment.hpp
```diff
@@ -114,21 +111,21 @@ class GlobalAssignment
         sequence<gpu>(o1.start, numPart, reorderFunctor.getBuf(), growthRate_);
         sortByKey<gpu>(keyView, std::span{reorderFunctor.getMap() + o1.start, keyView.size()}, s0, s1, growthRate_);
 
-        updateOctreeGlobal<KeyType>(keyView, bucketSize_, tree_, leaves_, d_csTree_, nodeCounts_, d_nodeCounts_);
+        updateOctreeGlobal<KeyType>(keyView, bucketSize_, tree_, leaves_, d_csTree_, nodeCounts_, d_nodeCounts_, false);
```
fthaler, line 114: Would it make more sense to put this line into an else block below or does it make to first attempt a first update without overflow check even on the first call?

fthaler (changes requested):

domain/include/cstone/domain/assignment.hpp
sekelle: It doesn't matter, the full convergence will make sure the tree is properly built on the first call. There's no errors or corruption in case overflow happens, just potentially some missed refinements that won't happen.

domain/include/cstone/tree/update_mpi_gpu.cuh
sekelle: I wouldn't use a compiler builtin. A plain `std::max` does the job and is more readable. The `std::numeric_limit` wouldn't give any gains here. Either number, `numeric_limit` or max(a,b), will be above the threshold that triggers cell refinement and a recount.

We could do
```c++
inout[i] = a + b >= std::max(a, b) ? a + b : std::numeric_limits<unsigned>::max();
```
but only if you think it's more readable.

fthaler: What I like about the builtin is that is contains ‘overflow’, so it makes it somewhat clear why `summax` is actually required. Otherwise, I would add a comment somewhere of why we use the custom reduction OP.

If there’s no comment, I guess `inout[i] = a + b >= std::max(a, b) ? a + b : std::numeric_limits<unsigned>::max();` is indeed more readable, otherwise it does not really matter I guess.

fthaler: Renaming `firstCall` to something verbose like `expectNodeCountOverflows` or similar might actually also help understanding the code.

sekelle: I've added these changes and comments.

fthaler (approved):

PR #553, opened 2025-11-25
Guard SPH against neighbor convergence failure

sekelle:
Particles that failed the smoothing length update, i.e. failed to keep neighbor counts within [ngmin, ngmax]
will be removed.

Currently this is active in the STD hydro propagator, and disabled in the VE propagator variants.
For now this is enough to do further tests on the TDE simulations.

A nicer interface with a controllable setting and an exception-based mechanism to write a checkpoint before failing
is in the works.

### DIFF
--- a/main/src/propagator/std_hydro.hpp
+++ b/main/src/propagator/std_hydro.hpp
@@ -181,7 +181,11 @@ public:
         computeTimestep(first, last, d);
         timer.step("Timestep");
         computePositions(groups_.view(), d, domain.box(), d.minDt, {float(d.minDt_m1)});
-        updateSmoothingLength(groups_.view(), d);
+        bool haveUnconvergedParticles = updateSmoothingLength(groups_.view(), d);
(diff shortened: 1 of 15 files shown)


fthaler (approved):

rmcabezon (changes requested):
This should not be a default behavior. This should be an option that the user can decide to apply or not. Lack of h/nc convergence can happen also in cases when the particle is not ejected from the system. So this cannot be a default behavior.

sekelle:
@rmcabezon , I added an option that can be used to control it. Default is still off.

rmcabezon (approved):

fthaler:
Just to make sure this is OK: the current changes still set `h[i]` to zero in the non-converged case while previously it was just kept on its last (non-converged) value. Is this change fine @rmcabezon?

sekelle:
Well the previous version immediately failed as soon as one particle didn't converge. This PR just slightly delays program abort in this case.

PR #557, opened 2025-12-12
Subset tagging frontend version 2 

ChristopherBignamini:
New implementation of the tagging user interface: tagging parameters are no longer stored (see #545). They are instead retrieved from the parameter file when needed. This pull request includes only parameter retrieval and validation. A follow-up PR will implement particle finding and output generation. The sphexa.cpp main function currently contains commented lines that act as placeholders for calls to the subset initialization and output functions. Same for the commented code block in ISimInitializer runTagging  method.

The parameters which control the tagging are currently the following :

- w_subset (same meaning of the existing -w argument but for id subsets)
- wextra_subset (same meaning of the existing --wextra but for id subsets)
- o_subset (same meaning of the existing -o but for id subsets)
- f_subset (same meaning of the existing -f but for id subsets)
- id_selection_spheres: list of comma separated floating points representing the sphere center coordinates and (non negative) radius (e.g., x_1, y_1, z_1, r_1, x_2, y_2, z_2, r_2, x_3, y_3, z_3, r_3 for 3 spheres)
- id_selection_spheres_group_ids: list of comma separated non negative integers, to specify the group of each sphere of the previous list (optional, if provided it must have the same size of id_selection_spheres)
- id_selection_list: list of comma separated non negative integers to tag specific ids
- id_selection_list_group_ids: list of comma separated non negative integers, to specify the group of each tagged id of the previous list (optional, if provided it must have the same size of id_selection_list)
- I have updated the set_params.py script to handle the new parameters, an example of call to add the above info to a parameter file is the following:

python set_parms.py tagging.h5 --add id_selection_spheres 0.,0.,0.,0.5,0.5,0.5,0.5,0.25,-0.5,-0.5,-0.5,0.25 id_selection_spheres_group_ids 0,1,2 id_selection_list 10,11,12 id_selection_list_group_ids 3,3,3

python set_parms.py tagging.h5 --add f_subset x,y,z,rho,vx,vy,vz --add o_subset subset_output --add w_subset 1 --add wextra_subset 1,10,0.77


main/src/sphexa/sphexa.cpp
```diff
@@ -134,7 +141,7 @@ int main(int argc, char** argv)
     float theta    = parser.get("--theta", haveGrav ? 0.5f : 1.0f);
 
     if (!parser.exists("-o")) { outFile += fileWriter->suffix(); }
-    if (writeEnabled) { writeSettings(simInit->constants(), outFile, fileWriter.get()); }
+    if (writeEnabled) { writeSettings(simInit->constants(), simInit->taggingSetup(), taggingOutputSetup, outFile, fileWriter.get()); }
```
sekelle, line 144: For saving the input tagging settings in the output, we can transfer these settings to the subset output file as file attributes. That simplifies things a lot, e.g. we don't need to store the spheres in the initializer and add an interface to get them out.

sekelle (changes requested):

PR #561, opened 2026-01-10
Update non-grav LET until converged

sekelle:
* The LET for the non-gravity case also needs to be checked for convergence as we no longer enforce the SFC-key boundaries of peer ranks.

* Improved the `RandomCoordinates` and `RandomGaussianCoordinates` sample sets with h-value estimation based on the distance of nearby particles along the space-filling-curve. Added an option to adjust h for neighbor counts within a given range based on the SFC estimates and added an option to randomly shuffle particles to destroy the SFC order.

* Use the improved sample coordinates with reasonable h values in domain tests to provide better coverage for halo particle detection and exchanges.

### DIFF
--- a/domain/include/cstone/domain/domain.hpp
+++ b/domain/include/cstone/domain/domain.hpp
@@ -188,15 +188,27 @@ public:
             focusTree_.converge(box(), keyView, global_.assignment(), global_.treeLeaves(), global_.nodeCounts(),
                                 invThetaEff, std::get<0>(scratch));
         }
-        focusTree_.updateMinMac(global_.assignment(), invThetaEff, true);
-        focusTree_.updateTree(global_.assignment(), global_.treeLeaves(), box(), std::get<0>(scratch));
-        focusTree_.updateCounts(keyView, global_.treeLeaves(), global_.nodeCounts(), std::get<0>(scratch));
 
-        reallocate(focusTree_.octreeViewAcc().numLeafNodes + 1, allocGrowthRate_, layout_, layoutAcc_);
-        focusTree_.discoverHalos(rawPtr(x), rawPtr(y), rawPtr(z), rawPtr(h), {rawPtr(layoutAcc_), layoutAcc_.size()},
-                                 haloSearchExt_, get<0>(scratch), false);
-        focusTree_.computeLayout({rawPtr(layoutAcc_), layoutAcc_.size()}, layout_);
-        halos_.exchangeRequests(focusTree_.treeLeaves(), focusTree_.assignment(), layout_);
+        int fail = 0, maxRep = 10;
+        do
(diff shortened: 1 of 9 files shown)


domain/include/cstone/domain/domain.hpp
```diff
@@ -188,15 +188,27 @@ class Domain
...
+                                     {rawPtr(layoutAcc_), layoutAcc_.size()}, haloSearchExt_, get<0>(scratch), false);
+            fail = focusTree_.computeLayout({rawPtr(layoutAcc_), layoutAcc_.size()}, layout_);
+            MPI_Allreduce(MPI_IN_PLACE, &fail, 1, MPI_INT, MPI_SUM, MPI_COMM_WORLD);
+
+            halos_.exchangeRequests(focusTree_.treeLeaves(), focusTree_.assignment(), layout_);
+
+            if (fail)
+            {
+                if (myRank_ == 0) { std::cout << "LET refine, mode=" << fail << std::endl; }
+            }
```
fthaler:
```suggestion
            if (fail && myRank == 0)
            {
                std::cout << "LET refine, mode=" << fail << std::endl;
            }
```

domain/test/coord_samples/random.hpp
```diff
@@ -263,11 +122,141 @@ void adjustSmoothingLength(LocalIndex numParticles,
...
+
+        for (std::size_t i = 0; i < n; ++i)
+        {
+            std::size_t j = i > hOffset ? i - hOffset : std::min(n, i + hOffset);
+
+            Vec3<T> pi{x_[i], y_[i], z_[i]};
+            Vec3<T> pj{x_[j], y_[j], z_[j]};
+
+            h_[i] = 0.5001 * std::sqrt(norm2(pi - pj));
+        }
```
fthaler: This is kind of unrelated to SFC sorting, isn’t it? At least the method name `sfcSort` wouldn’t make me expect that `h_` values get updated. So probably makes sense to either rename the method or factor out the update of `h_`.

fthaler (changes requested):

domain/test/coord_samples/random.hpp
sekelle: Was meant to be part of construction, so could be renamed and removed from the public interface. Coord sets are SFC sorted on construction by default, and can be scrambled upon request.

domain/include/cstone/domain/domain.hpp
sekelle: done

domain/test/coord_samples/random.hpp
sekelle: done

fthaler (approved):

PR #565, opened 2026-01-13
add flop counting to hydro_ve

OsmanSeckinSimsek:
adds a flop counter to the computeForces function in hydro_ve covering XMass, IAD, EOS, momentumEnergy calculations and saves the total FLOPs into the profiling hdf5 file.

### DIFF
--- a/main/src/propagator/hydro_flops.hpp
+++ b/main/src/propagator/hydro_flops.hpp
@@ -1,0 +1,233 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2021 CSCS, ETH Zurich
+ *               2021 University of Basel
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


sekelle (changes requested):
The flop counts computed here need a) the number of assigned (i) particles per rank and b) number of i-j pair interactions per rank. With these two numbers, the flop counts can be put together in e.g. a python script that reads the profiling output file.

We already have the number of i-particles. You can add the number of i-j pairs on top. The sum of all neighbor counts gets computed in the observables section where total energy is also computed. You could take the value from there and save it to the profiling output.

sekelle:
@OsmanSeckinSimsek, can you close this? All the metrics that you are aiming to compute can be deduced from the number of i and i-j pairs per rank. We already have the number of i particles in the profiling output and we compute the number of i-j pairs, but don't add it to the profiling output yet. So for computing these metrics, doing that (can be a new PR I guess) would be the way to go.

OsmanSeckinSimsek:
Sure, I'll close this and open a new PR adding i-j pairs to the profiling output.

PR #566, opened 2026-01-14
Initializers updated according to new particle data fields management

ChristopherBignamini:


main/src/init/file_init.hpp
```diff
@@ -51,8 +52,11 @@ void restoreDataset(IFileReader* reader, Dataset& d)
         {
             if (reader->rank() == 0) { std::cout << "restoring " << d.fieldNames[i]; }
             auto t0 = std::chrono::high_resolution_clock::now();
-            std::visit([reader, key = d.fieldNames[i]](auto field)
-                       { reader->readField(Dataset::prefix + key, field->data()); }, fieldPointers[i]);
+            std::visit([reader, key = d.fieldNames[i]](auto field){
+                auto tmp = toHost(*field);
```
sekelle: This will download an uninitialized array from the GPU. Not that the performance penalty will matter much compared to reading from disk, but it doesn't really show what's going on and might be a bit confusing for a reader.

Better to declare a temporary vector from `field->size()` with type `using T = decltype(*field->data())`.

main/src/init/file_init.hpp
```diff
@@ -180,14 +185,17 @@ class FileSplitInit : public ISimInitializer<Dataset>
             gatherSwap(y0, sfcOrder);
             gatherSwap(z0, sfcOrder);
 
+            auto x = toHost(d.x);
+            auto y = toHost(d.y);
+            auto z = toHost(d.z);
```
sekelle: No need to download these. Declare `x,y,z` as empty vectors.

main/src/init/file_init.hpp
```diff
@@ -221,28 +232,47 @@ class FileSplitInit : public ISimInitializer<Dataset>
...
-
+        auto m = toHost(d.m);
+        auto h = toHost(d.h);
+        auto vx = toHost(d.vx);
+        auto vy = toHost(d.vy);
+        auto vz = toHost(d.vz);
+        auto x_m1 = toHost(d.x_m1);
+        auto y_m1 = toHost(d.y_m1);
+        auto z_m1 = toHost(d.z_m1);
+        auto temp = toHost(d.temp);
```
sekelle: Avoid downloading uninitialized fields.

main/src/init/file_init.hpp
```diff
@@ -221,28 +232,47 @@ class FileSplitInit : public ISimInitializer<Dataset>
...
+        replicateField(reader, "m", m, T(1) / numSplits);
+        replicateField(reader, "h", h, T(1) / std::cbrt(numSplits));
+        replicateField(reader, "vx", vx, T(1));
+        replicateField(reader, "vy", vy, T(1));
+        replicateField(reader, "vz", vz, T(1));
+        replicateField(reader, "temp", temp, T(1));
```
sekelle: Adapt `replicateField` so it does the right thing when supplied with `d.m` etc

sekelle: `replicateField` can get the `dest = std::move(tmp)` at the end, do the replication into `tmp`.

main/src/init/file_init.hpp
```diff
@@ -221,28 +232,47 @@ class FileSplitInit : public ISimInitializer<Dataset>
...
+
+        d.m = std::move(m);
+        d.h = std::move(h);
+        d.vx = std::move(vx);
+        d.vy = std::move(vy);
+        d.vz = std::move(vz);
+        d.x_m1 = std::move(x_m1);
+        d.y_m1 = std::move(y_m1);
+        d.z_m1 = std::move(z_m1);
+        d.temp = std::move(temp);
```
sekelle: Should not be needed.

`std::transform` can be done directly on the GPU:
* Extend `scaleGpu` from primitives_gpu.cu, to have signature `(const T1* in1, const const T1* in2, T2* out, T2 value)
* Add `scaleGpuAcc` in primitives_acc.hpp that calls std::transform on CPU and `scaleGpu` on GPU
* Use `scaleGpuAcc` here instead of `std::transform`

main/src/init/isobaric_cube_init.hpp
```diff
@@ -82,55 +84,71 @@ void initIsobaricCubeFields(Dataset& d, const std::map<std::string, double>& con
...
+    auto&& x = toHost(d.x);
+    auto&& y = toHost(d.y);
+    auto&& z = toHost(d.z);
+    auto&& vx = toHost(d.vx);
+    auto&& vy = toHost(d.vy);
+    auto&& vz = toHost(d.vz);
+    auto h = toHost(d.h);
+    auto x_m1 = toHost(d.x_m1);
+    auto y_m1 = toHost(d.y_m1);
+    auto z_m1 = toHost(d.z_m1);
```
sekelle: Only `x,y,z` have valid data, don't download the other uninitialized fields.

sekelle (changes requested):
Applies to several initializers: don't use `toHost` to create vectors from uninitialized fields. Not primarily because of the performance impact, but because a reader might get the impression that there's a data-dependency on GPU data when in reality there's none.

sekelle (approved):
Some minor issue with u/temp fields usage which can be addressed later.

PR #569, opened 2026-01-21
Multiple init constants output test

ChristopherBignamini:


ci/cuda12/run_debug_sm90.yml
```diff
@@ -180,3 +180,107 @@ sph-run-cuda12-sm90-debug-sphexa-cpu:
...
+          fi
+        fi
+        wait
+      done
+      for ic in gresho-chan wind-shock kelvin-helmholtz; do
+        if [ "$SLURM_PROCID" -eq 0 ]; then
+          echo "Running test for init condition: $ic"
+        fi
+        wait
+        $APP_INSTALL_DIR/sphexa-cuda --quiet --glass ./50c.h5 --init $ic -s 2 -n 30
```
sekelle: * On GH200 we can use `-s 10 -n 50` for all IC cases, such that only one for loop will be required.
* Can we put the resulting bash script with a customizable application binary path into a separate file so we don't have to duplicate it between CPU/GPU and debug/release?
* We should run at least on 4 ranks, otherwise we miss a lot of coverage.

sekelle (changes requested):

sekelle (approved):

PR #570, opened 2026-01-23
Fix ryoanji CMake config

sekelle:
The object library was missing linkage to cstone_gpu and therefore missing some compile definitions.


ChristopherBignamini (approved):
lgtm

PR #571, opened 2026-01-24
Expand domain.syncGrav test coverage

sekelle:
There were no tests previously' for `domain.syncGrav`, only indirectly in Ryoanji through multipole upsweep and force computations.

### DIFF
--- a/ryoanji/test/interface/global_forces_gpu.cpp
+++ b/ryoanji/test/interface/global_forces_gpu.cpp
@@ -25,6 +25,7 @@
 
 #include "ryoanji/interface/global_multipole.hpp"
 #include "ryoanji/interface/multipole_holder.cuh"
+#include "ryoanji/nbody/traversal_cpu.hpp"
 
 using namespace ryoanji;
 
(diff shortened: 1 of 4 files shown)


fthaler (approved):

fthaler:
Looks good, but CI has to be fixed.

ryoanji/test/interface/global_forces_gpu.cpp
```diff
@@ -149,23 +180,31 @@ static int multipoleHolderTest(int thisRank, int numRanks)
         std::vector<T> azRef = dl(d_azref.data() + firstGlobalIdx, d_azref.data() + lastGlobalIdx);
 
         double         potentialSumRef = 0;
-        std::vector<T> errors(ax.size());
+        std::vector<T> errors(ax.size()), errors_cpu(ax.size());
```
ChristopherBignamini, line 184: I'm probably missing something but I don't see a declaration for errors_cpu

ChristopherBignamini (approved):
LGTM once the compilation error reported in the CI pipeline is fixed, I just have a question about line 183 in global_forces_gpu.cpp‎

ryoanji/test/interface/global_forces_gpu.cpp
sekelle: Similar to `double a, b;`

ChristopherBignamini: Ops, my apologies I missed one parenthesis and I was reading it as 

std::vector<T> errors(ax.size(), errors_cpu(ax.size())); 

complete non-sense... 😅

PR #572, opened 2026-01-26
Init based constants testing selection

ChristopherBignamini:


ci/scripts/run_constants_comparison.sh
```diff
@@ -41,14 +41,63 @@ if [ "$rank_id" -eq 0 ]; then
...
+    kelvin-helmholtz)
+      # Colums for TimeEnergyGrowth observables: iteration, ttot, minDt, etot, ecin, eint, egrav, linmom, angmom, khgr
+      echo "6"
+      ;;
+    *)
+      echo ""
+      ;;
+  esac
+}
+
```
sekelle, line 55: Bash has arrays, you could put the test case names and columns to use in array each.

sekelle (changes requested):

ci/scripts/compare_constants.py
sekelle: You can use `np.loadtxt(path)`, and then perform all diffs and relative errors on whole matrices or whole columns with numpy.

sekelle (changes requested):

ChristopherBignamini:
@sekelle I’ve implemented the suggested change (by creating a dedicated Python virtual environment for NumPy). I also slightly revised the test workflow: the results from each individual init case are now collected and failures are reported only at the end, as a list of failing tests.

sekelle (approved):

PR #573, opened 2026-01-28
Gradh correction

rmcabezon:
I am correcting the gradh terms because they were too noisy and the source of the granulation in the density for turbulence. Only apparent when dissipation was eliminated (as it should) in smooth flows with the avclean.
I had to split the VE from the Gradh calculation because now gradh need the kx from the neighbors, so a sync in between was needed.
This was built upon a commit that didn't have the new consolidated particle data for CPU and GPUs. So there might be things to change.


sekelle:
I've modified the ordering of SPH kernels in the VE propagator, such that we need one fewer round of halo exchanges and to show that the IAD loop can now be fused with GradH.

main/src/propagator/ve_hydro.hpp
```diff
@@ -149,38 +149,39 @@ class HydroVeProp : public Propagator<DomainType, DataType>
...
-        release(d, "gradh", "az");
-        acquire(d, "divv", "curlv");
+        release(d, "ay", "az");
+        acquire(d, "divv", "gradh");
         computeIadDivvCurlv(groups_.view(), d, domain.box());
         d.minDtRho = rhoTimestep(first, last, d);
         timer.step("IadVelocityDivCurl");
 
-        domain.exchangeHalos(get<"c11", "c12", "c13", "c22", "c23", "c33", "divv">(d), get<"ax">(d), get<"keys">(d));
+        computeGradh(groups_.view(), d, domain.box());
```
rmcabezon: A small preference: If possible, I'd like to have gradh and compute EOS before IAD. Although it doesn't make any difference, conceptually it is more in line with the standard SPH workflow. Just move 163-167 to 159. It makes it easier for me to follow the workflow.

main/src/propagator/ve_hydro.hpp
```diff
@@ -149,38 +149,39 @@ class HydroVeProp : public Propagator<DomainType, DataType>
...
-        timer.step("EquationOfState");
-
-        domain.exchangeHalos(get<"vx", "vy", "vz", "prho", "c", "kx">(d), get<"ax">(d), get<"keys">(d));
+        computeVe(groups_.view(), d, domain.box());
+        timer.step("Generalized Volume Elements");
+        domain.exchangeHalos(get<"vx", "vy", "vz", "kx">(d), get<"ax">(d), get<"keys">(d));
         timer.step("mpi::synchronizeHalos");
 
-        release(d, "gradh", "az");
-        acquire(d, "divv", "curlv");
```
rmcabezon, line 164: We never acquire curlv in the new version. Is it allocated?

main/src/propagator/ve_hydro.hpp
```diff
@@ -149,38 +149,39 @@ class HydroVeProp : public Propagator<DomainType, DataType>
...
         timer.step("AVswitches");
 
         if (avClean)
         {
-            domain.exchangeHalos(get<"dV11", "dV12", "dV13", "dV22", "dV23", "dV33", "alpha">(d), get<"ax">(d),
+            domain.exchangeHalos(get<"dV11", "dV12", "dV13", "dV22", "dV23", "dV33", "prho", "alpha">(d), get<"ax">(d),
                                  get<"keys">(d));
         }
-        else { domain.exchangeHalos(std::tie(get<"alpha">(d)), get<"ax">(d), get<"keys">(d)); }
+        else { domain.exchangeHalos(get<"prho", "alpha">(d), get<"ax">(d), get<"keys">(d)); }
```
rmcabezon, line 178: Why not sync prho in L169 instead of having to maintain it in two branches?

main/src/propagator/ve_hydro_bdt.hpp
rmcabezon, line 1: This propagator has not been updated in the same way as ve. It has my changes, but is missing your new reordering.

rmcabezon (commented):
Just a few comments.

main/src/propagator/ve_hydro.hpp
sekelle: The `gradh` can/should be fused with the IAD loop and the EOS includes the computation of `prho` which requires `gradh`.

main/src/propagator/ve_hydro.hpp
sekelle: It's acquired in the output section if `curlv` is requested there. In the compute_force part it's currently not used. But can be enabled, should it be needed.

main/src/propagator/ve_hydro.hpp
sekelle: `prho` could be synced at L169, but this sync has 8 fields already. Halo exchanges are done fused across fields, and exchanging more fields at the same time will increase memory usage. Deferring to L181 avoids syncing 9 fields at once. It can also increase performance as message sizes with just one field at L181 might be latency bound, such that adding a 2nd field would be for free in this case.

main/src/propagator/ve_hydro_bdt.hpp
sekelle: I didn't mean to update the `ve_bdt` variant. I wanted to do the lay work, allowing you to finish `ve` by fusing gradh and IAD.

sekelle (approved):

PR #577, opened 2026-02-25
Mac OS compatibility with size_t

nknk567:


ChristopherBignamini:
I’m not sure whether @nknk567 is seeing the same issue, but on my Mac (26.2) CMAKE_OSX_SYSROOT is not automatically picked up correctly, and as a result the Sedov solution build fails at link time.
I noticed that (at least on my system) CMake defines the APPLE variable automatically, but I’m not sure this is something we should rely on.
I’d suggest adding a note to the README recommending to manually set CMAKE_OSX_SYSROOT when running CMake (or via the ccmake UI under the compilation/configuration section).

nknk567:
> I’m not sure whether @nknk567 is seeing the same issue, but on my Mac (26.2) CMAKE_OSX_SYSROOT is not automatically picked up correctly, and as a result the Sedov solution build fails at link time. I noticed that (at least on my system) CMake defines the APPLE variable automatically, but I’m not sure this is something we should rely on. I’d suggest adding a note to the README recommending to manually set CMAKE_OSX_SYSROOT when running CMake (or via the ccmake UI under the compilation/configuration section).

Sedov solution builds without problem for me.

sekelle (approved):
Looks good, added some simplifications.

PR #578, opened 2026-02-25
Report number of unconverged particles

nknk567:


sekelle:
Hi Noah, you could compute this in the observables section where total energy, numbers of neighbors, etc is computed. Just add the local number of particles to the quantities to be reduced, and print a message if different from the previous global count.

main/src/observables/conserved_quantities.hpp
```diff
@@ -171,9 +167,10 @@ void computeConservedQuantities(size_t startIndex, size_t endIndex, Dataset& d,
...
     util::array<double, 3> globalLinmom{globalQuantities[3], globalQuantities[4], globalQuantities[5]};
     util::array<double, 3> globalAngmom{globalQuantities[6], globalQuantities[7], globalQuantities[8]};
-    d.linmom         = std::sqrt(norm2(globalLinmom));
-    d.angmom         = std::sqrt(norm2(globalAngmom));
-    d.totalNeighbors = size_t(globalQuantities[9]);
+    d.linmom             = std::sqrt(norm2(globalLinmom));
+    d.angmom             = std::sqrt(norm2(globalAngmom));
+    d.totalNeighbors     = size_t(globalQuantities[9]);
+    d.numParticlesGlobal = size_t(globalQuantities[10]);
 }
```
sekelle, line 175: You could preserve the number of particles from the previous iteration in `d.numParticlesGlobalPrev` here.

main/src/observables/time_energies.hpp
```diff
@@ -53,12 +53,19 @@ class TimeAndEnergy : public IObservables<Dataset>
...
         if (rank == 0)
         {
             fileutils::writeColumns(constantsFile, ' ', d.iteration, d.ttot, d.minDt, d.etot, d.ecin, d.eint, d.egrav,
                                     d.linmom, d.angmom);
+            if (numParticlesGlobal_previous != d.numParticlesGlobal)
+            {
+                std::cout << "particle number changed: "
+                          << std::int64_t(d.numParticlesGlobal) - std::int64_t(numParticlesGlobal_previous)
+                          << std::endl;
+            }
```
sekelle: If you store the previous number of particles, you can print the difference in `printIterationTimings` together with all the other metrics. Those belong thematically into the same group and I would avoid printing them from different places.

main/src/observables/conserved_quantities.hpp
sekelle: `d.numParticlesGlobalPrev` would not have to be preserved for IO in `loadOrStore`

sekelle (changes requested):

main/src/sphexa/sphexa.cpp
```diff
@@ -121,7 +121,8 @@ int main(int argc, char** argv)
     propagator->load(initCond, fileReader.get());
     auto box = simInit->init(rank, numRanks, problemSize, simData, fileReader.get());
 
-    auto& d = simData.hydro;
+    auto& d                  = simData.hydro;
+    d.numParticlesGlobalPrev = d.numParticlesGlobal;
```
sekelle, line 125: This is assignment is not needed, the observable calculation will set it in any case.

sekelle (approved):

PR #582, opened 2026-03-15
CI job script refactoring

ChristopherBignamini:


ChristopherBignamini:
I’m investigating the origin of the failures in the constant comparison tests when restarting jobs. The discrepancies are small, but the tests were passing before the CI script refactoring. The failures disappear when running with a single MPI task, so it is likely just a matter of numerical noise. #572 will fix this issue, with a better control on check handling. In the meantime, I've increased the threshold for absolute discrepancy check selection wrt relative one.

sekelle (approved):

PR #583, opened 2026-03-17
Fixes for CUDA 13 and OpenMPI

sekelle:


domain/include/cstone/primitives/mpi_wrappers.hpp
```diff
@@ -22,19 +22,25 @@
 
 #include "cstone/util/type_list.hpp"
 
+namespace detail
+{
+// clang-format off
+// nvcc doesn't allow making this constexpr with OpenMPI, as the type macros contain a static_cast<void*>
```
fthaler, line 28: Did you check if the original code works with `--expt-relaxed-constexpr` (e.g., adding `add_compile_options($<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:--expt-relaxed-constexpr>)` to CMakeLists.txt)? Clustered neighbor lists require this anyway, so could be worth a try.

main/src/propagator/ipropagator.hpp
```diff
@@ -104,23 +104,24 @@ class Propagator
...
-        auto nodeCount          = domain.globalTree().numLeafNodes;
-        auto particleCount      = domain.nParticles();
-        auto haloCount          = d.maxHalos;
-        auto totalNeighbors     = d.totalNeighbors;
-        auto totalParticleCount = d.numParticlesGlobal;
+        auto nodeCount        = domain.globalTree().numLeafNodes;
+        auto particleCount    = domain.nParticles();
+        auto haloCount        = d.maxHalos;
+        auto totalNeighbors   = d.totalNeighbors;
+        auto avgNcPerParticle = rank_ == 0 ? totalNeighbors / d.numParticlesGlobal : 0;
```
fthaler: Why is the `rank_ == 0` check required?

domain/include/cstone/primitives/mpi_wrappers.hpp
sekelle: No, doesn't work, the OpenMPI type macros expand to:
```c++
(static_cast<MPI_Datatype>(static_cast<void*>(&(ompi_mpi_double))))
```
That's more or less a manual attempt to prevent it from being a compile time constant, rather than a limitation in nvcc. GCC just seems happy to eliminate the superfluous double static_cast.

main/src/propagator/ipropagator.hpp
sekelle: Explained why and made this explicit in the code - global number of particles are only valid on rank 0.

domain/include/cstone/primitives/mpi_wrappers.hpp
fthaler: I hoped NVCC would accept the `static_cast` as `constexpr` but nevermind then.

fthaler (approved):

jgphpc (approved):

PR #584, opened 2026-03-20
Explicitly write step attribute value provided by rank 0

sekelle:
When opening a file for parallel writes during checkpointing, `stepAttribute` nevertheless has to be called on all ranks,
even though calling with different attribute values constitutes a data race.

As a work-around, I broadcast the attribute of rank 0 to all other ranks to support the case
where only rank 0 has a valid attribute value.

### DIFF
--- a/main/src/io/ifile_io_hdf5.cpp
+++ b/main/src/io/ifile_io_hdf5.cpp
@@ -103,7 +103,10 @@ public:
                 // so we explicitly broadcast the value of rank 0 to all ranks
                 using Type = std::decay_t<decltype(*argp)>;
                 std::vector<Type> arg(size);
-                std::copy(argp, argp + size, arg.data());
+                if (size > 0 && rank_ == 0)
+                {
+                    std::copy(argp, argp + size, arg.data());
+                }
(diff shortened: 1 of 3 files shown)


main/src/io/ifile_io_hdf5.cpp
```diff
@@ -208,7 +214,14 @@ class H5PartWriterSeq final : public H5PartWriter
...
     {
-        if (rank_ == 0) { Base::stepAttribute(key, val, size); }
+        if (rank_ == 0)
+        {
+            std::visit(
+                [this, &key, size](auto arg) { //
+                    fileutils::H5WriteStepAttribT(h5File_, key.c_str(), arg, size);
+                },
+                val);
+        }
```
sekelle, line 227: Base version no longer compatible, so I have to implement it here.

rmcabezon (approved):
Nice to use Copilot to review.

biddisco:
LGTM : The basic idea of broadcasting the attribute before write is sound, but probably some of copilot's suggestions should be addressed before merging.

PR #592, opened 2026-05-04
visualization: refactoring and improvements

albestro:
WIP

Changelog:
- isolate visualization stuff in `viz` folder and related CMake target
- switch from C-like API to RAII for MPI initialization with `MPIScope`, and also for `CatalystAdaptor` and `AscentAdaptor`
- factor out mesh-blueprint conduit node setup, so it is shared by all adaptors
- use `conduit_cpp` in `CatalystAdaptor` and `conduit` in `AscentAdaptor`

TODO:
- [x] document `define_mesh` template solution (different semantic among conduit_cpp and conduit)
- [x] refine implicit connectivity (can be dropped with next ascent release?)
- [x] cleanup unused/commented fields
- [x] test and document possible setups and their (current?) limits


sekelle (approved):
lgtm

PR #597, opened 2026-05-22
Slr

rmcabezon:
- Added the new SLR dissipation that replaces the old avclean
- the --avclean is not needed anymore. SLR is the default if no arguments are provided.
- It is possible to deactivate SLR with --no-slr
- AV switches are deactivated by default. They can be activated with --avswitches
- In principle, it is possible to run with no AV if --no-slr is used w/o using --avswitches. We could cap this and not allow it, but maybe we would like to do a test in this respect in the future and would like to keep the flexibility
- An additional limiting floor for the dissipation can be used following the formalism of Garcia-senz and Cabezon (2026). It can activate with --avfloor <N>, where <N> is a real number. Default is deactivated with N=1. Recommended value for subsonic turbulence only (so far) N=0.5. In principle, N should be bracketed between [0,1], but it is not at the moment.
- Reduced the stEnergyPrefac for turbulence test case so that the RMS(c)=0.3 and not 0.35 as it was until now.


sph/include/sph/hydro_ve/momentum_energy_kern.hpp
```diff
@@ -113,18 +113,32 @@ struct MomentumAndEnergyInteraction
...
+            T balsj   = T(1) - f_j * f_j;
+            rv_slr += avRvCorrection({rx, ry, rz}, stl::min(v1, v2), eta_crit, balsi, balsj,
+                                       {dV11i, dV12i, dV13i, dV22i, dV23i, dV33i},
+                                       {dV11j, dV12j, dV13j, dV22j, dV23j, dV33j});
         }
 
         T wij          = i == j ? 0 : rv / dist;
-        T viscosity_ij = artificial_viscosity(alpha_i, alpha_j, ci, cj, wij);
+        T wij_slr      = i == j ? 0 : rv_slr / dist;
+        T viscosity_ij = artificial_viscosity(alpha_i, alpha_j, ci, cj, wij_slr, wij_slr, Lij);
```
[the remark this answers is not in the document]
rmcabezon: @copilot It is not unintended. Initially, we explored the possibility of having wij_slr for the linear part of the AV and wij for the quadratic part, but we saw that it worked better to have wij_slr in both parts. Admittedly, I could remove the wij argument and pass only wij_slr, but I want to conserve this flexibility for a while.

sph/test/ve.cpp
```diff
@@ -100,12 +100,13 @@ class SphKernelTests : public testing::Test
     T Atmin          = 0.1;
     T Atmax          = 0.2;
     T ramp           = 1.0 / (Atmax - Atmin);
+    T avFloor        = 1.0;
 
     uint64_t                        npart          = 99;
```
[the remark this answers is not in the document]
rmcabezon: Done

main/src/sphexa/sphexa.cpp
```diff
@@ -136,6 +139,19 @@ int main(int argc, char** argv)
...
+        d.alphamin                = 0.05;
+        if (resetAlpha) { cstone::fill(exec, d.alpha.begin(), d.alpha.end(), d.alphamin); }
+    }
+    if (haveAvFloor) { d.avFloor = parser.get<double>("--avfloor"); }
+    if (!AVswitches && initFromFile)
+    {
+        d.alphamin         = 1.0;
+        d.alphamax         = 1.0;
+        cstone::fill(exec, d.alpha.begin(), d.alpha.end(), d.alphamin);
+    }
```
sekelle: This logic initialized `alpha`, hence it should be part of the initalizer.

sph/include/sph/hydro_ve/momentum_energy_kern.hpp
```diff
@@ -113,18 +113,32 @@ struct MomentumAndEnergyInteraction
...
+        if constexpr (SLR)
         {
-            rv += avRvCorrection({rx, ry, rz}, stl::min(v1, v2), eta_crit, {dV11i, dV12i, dV13i, dV22i, dV23i, dV33i},
-                                 {dV11j, dV12j, dV13j, dV22j, dV23j, dV33j});
+            T eps_i   = T(1e-4) * ci * hiInv;
+            T eps_j   = T(1e-4) * cj * hjInv;
+            T denom_i = std::abs(divvi) + std::abs(curlvi) + eps_i;
+            T denom_j = std::abs(divvj) + std::abs(curlvj) + eps_j;
+            T f_i     = (denom_i > T(0)) ? std::abs(divvi) / denom_i : T(0);
+            T f_j     = (denom_j > T(0)) ? std::abs(divvj) / denom_j : T(0);
```
sekelle: The denominators add a non-zero eps term, it seems like they can never be zero and we can remove the check?

main/src/sphexa/sphexa.cpp
sekelle: After a 2nd look, this logic is not needed.

* You're starting fresh, no init file -> `alpha` will be set to `alphamin` as specified by the initializer (sedov, turbulence, ...)

* You're starting from file -> `alpha` will be filled from the file contents. If you want to change the alphas in the file, change the values with a python + h5py script.

sekelle:
@rmcabezon PR is good to merge from my side now, only one small issue:
I'm computing `divv` from the velocity jacobian, but the SPH test example data has one line where divv != dVxx + dVyy + dVzz, hence I think the code is fine, but the test reference should be updated.

sph/include/sph/hydro_ve/momentum_energy_kern.hpp
rmcabezon: Yes.

sph/include/sph/hydro_ve/momentum_energy_kern.hpp
```diff
@@ -121,16 +121,18 @@ struct MomentumAndEnergyInteraction
         {
             T eps_i   = T(1e-4) * ci * hiInv;
             T eps_j   = T(1e-4) * cj * hjInv;
-            T denom_i = std::abs(divvi) + std::abs(curlvi) + eps_i;
-            T denom_j = std::abs(divvj) + std::abs(curlvj) + eps_j;
-            T f_i     = (denom_i > T(0)) ? std::abs(divvi) / denom_i : T(0);
-            T f_j     = (denom_j > T(0)) ? std::abs(divvj) / denom_j : T(0);
+            T divvi   = std::abs(dV11i + dV22i + dV33i);
+            T divvj   = std::abs(dV11j + dV22j + dV33j);
+            T denom_i = divvi + std::abs(curlvi) + eps_i;
```
rmcabezon, line 126: Conceptual change: Maybe better apply the std::abs  to divvi when the denom is calculated and not to the calculation of divvi/j. This is in case we want to reuse divvi in this file, so that we don't inadvertently include the abs, when it is not needed.

sph/include/sph/hydro_ve/momentum_energy_kern.hpp
```diff
@@ -135,7 +135,6 @@ struct MomentumAndEnergyInteraction
                                {dV11i, dV12i, dV13i, dV22i, dV23i, dV33i}, {dV11j, dV12j, dV13j, dV22j, dV23j, dV33j});
         }
 
-        T wij          = i == j ? 0 : rv / dist;
```
rmcabezon, line 138: If we remove this variable, then you can simplify the call for artificial_viscosity function to 
artificial_viscosity(alpha_i, alpha_j, ci, cj, wij_slr, Lij); and use the same wij inside of that function, everywhere.

sph/include/sph/hydro_std/momentum_energy_kern.hpp
```diff
@@ -60,7 +60,7 @@ struct MomentumAndEnergyInteractionStd
 
         T           wij          = i == j ? 0 : rv / dist;
         constexpr T av_alpha     = T(1);
-        T           viscosity_ij = T(0.5) * artificial_viscosity(av_alpha, av_alpha, ci, cj, wij);
+        T           viscosity_ij = T(0.5) * artificial_viscosity(av_alpha, av_alpha, ci, cj, wij, wij, T(1));
```
rmcabezon, line 63: Remove one wij from the interface.

sph/include/sph/hydro_ve/momentum_energy_kern.hpp
```diff
@@ -113,18 +113,33 @@ struct MomentumAndEnergyInteraction
...
+            T balsj   = T(1) - f_j * f_j;
+            rv_slr +=
+                avRvCorrection({rx, ry, rz}, stl::min(v1, v2), eta_crit, balsi, balsj,
+                               {dV11i, dV12i, dV13i, dV22i, dV23i, dV33i}, {dV11j, dV12j, dV13j, dV22j, dV23j, dV33j});
         }
 
-        T wij          = i == j ? 0 : rv / dist;
-        T viscosity_ij = artificial_viscosity(alpha_i, alpha_j, ci, cj, wij);
+        T wij_slr      = i == j ? 0 : rv_slr / dist;
+        T viscosity_ij = artificial_viscosity(alpha_i, alpha_j, ci, cj, wij_slr, wij_slr, Lij);
```
rmcabezon, line 139: Remove one wij_slr from the artificial_viscosity interface.

sph/include/sph/kernels.hpp
```diff
@@ -60,25 +60,26 @@ HOST_DEVICE_FUN inline T wharmonic_derivative_std(T v)
...
+ * @param c_i      speed of sound particle i
+ * @param c_j      speed of sound particle j
+ * @param w_ij     relative velocity (v_i - v_j), projected onto the connecting axis (r_i - r_j)
+ * @param w_ij_slr relative SLR corrected velocity (v'_i - v'_j), projected onto the connecting axis (r_i - r_j)
+ * @param Lij      Floor limit for the linear term (Default: Lij=1.0, i.e. deactivated)
+ * @return         the viscosity
  */
 template<typename T>
-HOST_DEVICE_FUN inline T artificial_viscosity(T alpha_i, T alpha_j, T c_i, T c_j, T w_ij)
+HOST_DEVICE_FUN inline T artificial_viscosity(T alpha_i, T alpha_j, T c_i, T c_j, T w_ij, T w_ij_slr, T Lij)
```
rmcabezon, line 74: remove w_ij_slr

sph/include/sph/kernels.hpp
```diff
@@ -60,25 +60,26 @@ HOST_DEVICE_FUN inline T wharmonic_derivative_std(T v)
...
 {
-    // alpha is const for now, but will be different for each particle when using viscosity switching
     constexpr T beta = T(2.0);
 
     T viscosity_ij = T(0.0);
     if (w_ij < T(0.0))
     {
-        T vij_signal = (alpha_i + alpha_j) * T(0.25) * (c_i + c_j) - beta * w_ij;
-        viscosity_ij = -vij_signal * w_ij;
+        T vij_signal = Lij * (alpha_i + alpha_j) * T(0.25) * (c_i + c_j) * w_ij_slr - beta * w_ij * w_ij;
```
rmcabezon, line 81: Rename w_ij_slr as w_ij

PR #598, opened 2026-05-26
Use of mixed dimension (MixD) keys in SFC

iomaganaris:
This PR mixed precision keys that are based on the Hilbert SFC. The keys are generated using the following rule:
```
b31 b30 b29 b28 ... b16 b15 b14 ... b8  b7  b6  b5  b4  b3  b2  b1  b0
x   x   1Dk 1Dk ... 1Dk 2Dk 2Dk ... 2Dk 2Dk 3Dk 3Dk 3Dk 3Dk 3Dk 3Dk 3Dk
        <- 15 1D keys -><---- 8 2D keys ---><------- 7 3D keys ------->
```
To figure out how many 1D, 2D and 3D keys we need, we use some extra variables `bx`, `by`, `bz`. Those at the moment at automatically defined by the aspect rations of the lengths of the axes of the domain bounding box. This calculation is done in the `getBoxMixDimensionBits` function.
The range of bit encoding are defined by first looking at the longest length of the domain bounding box which gets the full range of the key, i.e. if `x` axis is the longest, `bx` will be `10` for `32 bit` keys and `21` for `64 bit` keys. Then the rest of the axes' dimensions are reduced by one bit for every factor-of-2 they are smaller than the longest.
For example for the following box:
```
Box<double>{0, 1, 0, 0.015625, 0, 0.00390625}
```
will have the following ranges:
```
KeyType    | maxLevel | bx | by | bz
-----------|----------|----|----|-—-
unsigned   |    10    | 10 |  4 |  2
uint64_t   |    21    | 21 | 15 | 13
```
The key is generated using the `iHilbertMixD` function. It encodes a 3D integer coordinate (px, py, pz) into a Hilbert space-filling curve key for anisotropic boxes, where each axis gets a different number of bits (bx, by, bz) proportional to its physical length. It handles the unequal bit counts by peeling off the "extra" bits of the longer dimensions first as 1D then 2D Hilbert segments, before encoding the remaining equal-depth cube with the standard 3D Hilbert curve.

The result of the MixD SFC keys is that particles that are closer together in space are grouped together in the same nodes of the octree for the non uniform length domains. Parts of the tree that are not valid keys of the MixD SFC keys have size `0` and are excluded from searches.

Performance for a disk-like domain where `x` and `y` dimensions are long and `z` is shorter can be seen below:
```
Run:     kelvin_r4_c1_s200_n50_gpu
MixD    numHalos sum : 206,683,489
Develop numHalos sum : 398,082,025
Ratio   MixD / Dev   : 0.519198

=== Timing Comparison: MixD vs develop  [KELVIN / GPU]  (summed over all steps) ===
─────────────────────────────────────────────────────────────────────────────────────────────────
Category                                          MixD (s)  Develop (s)     Diff (s)     Diff (%)
─────────────────────────────────────────────────────────────────────────────────────────────────
domain::sync                                         3.644        3.471       +0.173         +5.0
computeGroups                                        0.166        0.000       +0.166          n/a
updateSmoothingLengthIterative                       2.794        0.000       +2.794          n/a
FindNeighbors                                        3.789       15.077      -11.287        -74.9
XMass                                                1.493        4.218       -2.724        -64.6
mpi::synchronizeHalos                                0.980        2.005       -1.025        -51.1
Generalized Volume Elements                          1.441        3.402       -1.961        -57.6
IadVelocityDivCurlGradh                              8.548       19.869      -11.321        -57.0
EquationOfState                                      0.019        0.209       -0.191        -91.0
AVswitches                                           3.983        8.182       -4.199        -51.3
MomentumAndEnergy                                    8.325       19.680      -11.354        -57.7
Timestep                                             0.191        3.695       -3.504        -94.8
UpdateQuantities                                     0.059        0.325       -0.266        -81.8
Total time reported                                 39.428       90.598      -51.170        -56.5
─────────────────────────────────────────────────────────────────────────────────────────────────
```
The above timings are taken using 4 ranks and 4 GPUs in a single node.
To execute this specific test in 1 node using 4 GPUs you can use the `run_gpu.sh` script by editing the number of ranks and cores used in the script:
```
sbatch run_gpu.sh kelvin
```
This script executes the `Kelvin-Helmholtz` experiment.

TODOs
- [x] Improve PR description
- [x] Add GPU tests for MixD domains
- [x] Clean up files added to this PR
- [x] Add necessary documentation in new functions


sekelle:
Additional todos:

* Mixed-D type is already runtime dynamic, so we don't need a new KeyType -> remove `HilbertMixD` and replace `Hilbert` with the new one.

* `AxisMixDBits` -> `AxesBits` implemented as `util::array<int, 3>`, supply this to all functions now taking `unsigned bx, by, bz` and a box. (i.e. everywhere except `hilbert.hpp`, which does not include `box.hpp`)

* Replace old `containedIn` with the new version, remove duplicate.

* Replace old `sfcIBox` with new version taking `AxesBits`

sekelle:
* `sfcIBox(key, level, axesBits)` should take level counted from the root for consistency with the rest of the code.

sekelle:
* consolidate tests in `traversal.cpp` into a single version

scripts/analysis/compare_num_halos.py
sekelle, line 1: I think we can remove this one.

scripts/analysis/timing_summary.py
sekelle, line 1: I think this can be removed.

scripts/analysis/compare_sphexa_tests.py
sekelle, line 1: Should be removed, we have a version of this in the `ci` folder.

scripts/benchmark/run_common.sh
sekelle, line 1: Hmm I think we could collect these run scrips in a separate repo.

main/src/sphexa/sphexa.cpp
```diff
@@ -157,6 +157,33 @@ int main(int argc, char** argv)
...
+        }
+#else
+        // CPU: host vectors, write directly
+        for (size_t i = 0; i < count; ++i)
+        {
+            rankFile << d.x[start + i] << " " << d.y[start + i] << " " << d.z[start + i] << "\n";
+        }
+#endif
+    }
+
```
sekelle: Left over debug or experimental profiling output that should be removed.

domain/test/integration_mpi/domain_gpu.cpp
```diff
@@ -64,20 +63,38 @@ void randomGaussianAssignment(int rank, int numRanks)
...
     std::vector<T> hs1, hs2, hs3;
     domainCpu.sync(keys, x, y, z, h, std::tie(m, rungs), std::tie(hs1, hs2, hs3));
 
+    // Make sure that small box shrinking doesn't change the MixD bits compared to the original box
+    const auto cpuDomainBoxMixDBits = getBoxDimensionBits<T, KeyType, Box<T>>(domainCpu.box());
+    const auto isCpuDomainBoxMixD =
+        (cpuDomainBoxMixDBits[0] != maxTreeLevel<KeyType>{} || cpuDomainBoxMixDBits[1] != maxTreeLevel<KeyType>{} ||
+         cpuDomainBoxMixDBits[2] != maxTreeLevel<KeyType>{});
+    EXPECT_EQ(isOriginalBoxMixD, isCpuDomainBoxMixD);
+
```
sekelle: We don't need to check for this, we can prevent tight-fitting bounding boxes by setting the boundary type to `BoundaryType::fixed`

domain/include/cstone/focus/source_center.hpp
```diff
@@ -29,6 +29,8 @@ template<class Tc, class Th>
 HOST_DEVICE_FUN util::tuple<Vec3<Tc>, Vec3<Tc>> computeBoundingBox(
     const Tc* x, const Tc* y, const Tc* z, const Th* h, LocalIndex first, LocalIndex last, Th scale, Vec3<Tc> init)
 {
+    if (first == last) { return {Vec3<Tc>{0, 0, 0}, Vec3<Tc>{0, 0, 0}}; }
+
```
sekelle: This should be removed, the empty box case is handled by the `init` argument

sekelle (changes requested):

domain/include/cstone/focus/source_center.hpp
sekelle: done

domain/test/integration_mpi/domain_gpu.cpp
sekelle: done

main/src/sphexa/sphexa.cpp
sekelle: done

scripts/analysis/compare_sphexa_tests.py
sekelle, line 1: We have a comparison script in the ci directory, can this be used or made more flexible to support this use case instead of adding another?

scripts/analysis/compare_num_halos.py
```diff
@@ -0,0 +1,64 @@
...
+
+if args.cores is not None:
+    CORES = args.cores
+elif BACKEND == "cpu":
+    CORES = 18
+else:
+    CORES = 1
+
+MIXD_ROOT = "/capstor/scratch/cscs/ioannmag/CORNERSTONE/sphexa"
+DEV_ROOT  = "/capstor/scratch/cscs/ioannmag/CORNERSTONE/sphexa-develop"
```
sekelle: Hard coded path here

scripts/benchmark/run_common.sh
```diff
@@ -0,0 +1,126 @@
...
+for _arg in "$@"; do
+  [ "$_arg" = "--nsys" ] && NSYS_ENABLED=1
+done
+N=50
+STEPS=200
+
+# ── Directories ────────────────────────────────────────────────────────────────
+MIXD_ROOT="/capstor/scratch/cscs/ioannmag/CORNERSTONE/sphexa"
+DEV_ROOT="/capstor/scratch/cscs/ioannmag/CORNERSTONE/sphexa-develop"
+GLASS="/capstor/scratch/cscs/ioannmag/CORNERSTONE/50c.h5"
```
sekelle: hard coded path here

sekelle (approved):

PR #600, opened 2026-05-28
Conditioning treatment in IAD

nknk567:
Regularize IAD matrix if det(tau)/tr(tau)^3 falls below a certain quality target.

- quality target is set in ParticlesData as attribute (turned off by default)
- regularized particles are tagged in their id field, number of regularized particles can be printed in the propagator
- new file sph/id_layout.hpp specifies number of bits used for group tagging and the location of the bit to mark regularization such that they don't conflict
- printAndClearIadRegularizationStats "clear"-parameter can be turned off to have regularization information for each particle visible in the file output


rmcabezon:
Hi @nknk567. Is this ready for review? Maybe you can submit the PR?

sph/include/sph/iad_regularization_stat.cu
```diff
@@ -0,0 +1,36 @@
...
+    auto end   = id.data() + lastIndex;
+
+    auto              check_mask = [] __device__(std::uint64_t value) { return (value & iadRegularizationMask) != 0; };
+    const std::size_t count      = thrust::count_if(thrust::device, begin, end, check_mask);
+
+    if (clear)
+    {
+        auto clear_mask = [] __device__(std::uint64_t value) { return value & ~iadRegularizationMask; };
+        thrust::transform(thrust::device, begin, end, begin, clear_mask);
+    }
```
sekelle: There should be no clearing needed, because the IAD kernel already clears the regularization bit on each iteration.

nknk567: I included the clearing so it's not affecting the output files. We could do it in saveFields() instead.

sekelle (approved):

sekelle (approved):

PR #605, opened 2026-06-10
Support for GPU Streams in Cornerstone

fthaler:
Introduces GPU stream support inside cornerstone (the domain subdirectory of SPH-EXA).
- Provides `cstone::execution::cpu`, `cstone::execution::gpuDefaultStream`, and `cstone::execution::gpuStream(...)` execution policies.
- Uses these execution policies throughout the cornerstone library, removes any use of the default stream apart from uses within the `DeviceVector` class which still performs all memory operations on the default stream.
- Removes all `<bool useGpu>` template parameters on functions. Instead, uses overloads based on the execution policies.
- Removes duplicate functions in primitives/primitives_acc (e.g. `fillGpu` is now just `fill` and overloads to the correct implementation based on the execution policy).
- Moves some functions (i.e., MPI wrappers and `rawPtr`) into the `cstone` namespace as a byproduct of some required refactoring to simplify the CPU and GPU function overloads.


domain/include/cstone/primitives/concat_vector.hpp
```diff
@@ -58,28 +59,38 @@ class ConcatVector
...
+        memcpyD2HAsync(exec, src.data().data(), src.data().size(), dstBuffer);
     }
     else if constexpr (IsDeviceVector<AccVec1<T>>{} && IsDeviceVector<AccVec2<T>>{})
     {
-        memcpyD2D(src.data().data(), src.data().size(), dstBuffer);
+        memcpyD2DAsync(exec, src.data().data(), src.data().size(), dstBuffer);
     }
     else if constexpr (!IsDeviceVector<AccVec1<T>>{} && !IsDeviceVector<AccVec2<T>>{})
     {
         std::copy_n(src.data().data(), src.data().size(), dstBuffer);
```
sekelle: This case is identical to the separate `execution::Cpu` overload. If this is called with `execution::Gpu`, maybe we don't need the overload and can take a generic `exec` as a template?

domain/include/cstone/primitives/primitives_acc.hpp
```diff
@@ -17,107 +17,96 @@
...
-    auto s2                 = valueBuf.size();
-    uint64_t numElements[2] = {uint64_t(keys.size() * growthRate), tempStorageEle};
-    auto tempBuffers        = util::packAllocBuffer<ValueType>(valueBuf, {numElements, 2}, 128);
-
-    sortByKeyGpu(keys.data(), keys.data() + keys.size(), values.data(), (KeyType*)rawPtr(keyBuf), tempBuffers[0].data(),
-                 tempBuffers[1].data(), tempStorageEle * sizeof(ValueType));
-    reallocate(keyBuf, s1, 1.0);
-    reallocate(valueBuf, s2, 1.0);
+    assert(keys.size() == values.size());
+    sort_by_key(keys.begin(), keys.end(), values.begin());
```
sekelle, line 88: This routes all cases through the simple thrust interface that allocates key and value buffers instead of the direct CUB calls that reuses memory.

We should remove the simple thrust version, to avoid accidents like this one from happening in the future.

domain/include/cstone/traversal/ijloop/ijloop.hpp
```diff
@@ -118,25 +119,30 @@ struct ConceptTestInteraction
...
-                                       Box<double> box,
-                                       LocalIndex totalBodies,
-                                       GroupView groups,
-                                       const double* x,
-                                       const double* y,
-                                       const double* z,
-                                       const float* h)
+template<class T, class Exec>
+concept NeighborhoodBuilderWithExec = execution::Policy<Exec> && requires(Exec exec,
+                                                                          T nb,
```
sekelle: Is the `WithExec` suffix really needed?

domain/include/cstone/tree/csarray_gpu.cu
```diff
@@ -175,45 +180,59 @@ processNodes(const KeyType* oldTree, const TreeNodeIndex* nodeOps, TreeNodeIndex
...
+    rebalanceDecisionKernel<<<iceil(numNodes, nThreads), nThreads, 0, exec>>>(tree, counts, numNodes, bucketSize,
+                                                                              nodeOps);
 
     size_t nodeOpsSize = numNodes + 1;
-    thrust::exclusive_scan(thrust::device, nodeOps, nodeOps + nodeOpsSize, nodeOps);
+    thrust::exclusive_scan(thrustExecPolicy(exec), nodeOps, nodeOps + nodeOpsSize, nodeOps);
 
     TreeNodeIndex newNumNodes;
-    thrust::copy_n(thrust::device_pointer_cast(nodeOps) + nodeOpsSize - 1, 1, &newNumNodes);
+    thrust::copy_n(thrustExecPolicy(exec), thrust::device_pointer_cast(nodeOps) + nodeOpsSize - 1, 1, &newNumNodes);
```
sekelle: Specifying the GPU execution policy here is an error and will crash on systems without unified CPU-GPU memory. Remove the policy and it will do the right thing.

sph/include/sph/hydro_turb/driver.hpp
sekelle, line 1: Seems like a line ending change, is it justified, i.e. is the new version the correct one?

sekelle (changes requested):

sph/include/sph/hydro_turb/driver.hpp
fthaler: Correct. I created a separate PR which fixes all line ending issues here, probably makes more sense: https://github.com/sphexa-org/sphexa/pull/610

domain/include/cstone/primitives/primitives_acc.hpp
fthaler: This is the CPU path, no CUB involved.

domain/include/cstone/primitives/concat_vector.hpp
fthaler: I was assuming that this code path is still used somewhere as it existed before. But it is indeed not, so I just removed it now and added a `static_assert`. So we can deal with it if there is actually a use case for it.

domain/include/cstone/traversal/ijloop/ijloop.hpp
fthaler: It is required to distinguish it from the actualy `NeighborhoodBuilder` concept below. But probably I will wrap it inside the `detail` namespace as it’s not meant to be used anywhere else.

domain/include/cstone/tree/csarray_gpu.cu
fthaler: Good catch, fixed.

domain/include/cstone/primitives/primitives_acc.hpp
sekelle: Right, missed it. Nevertheless, we can remove the simple version without arguments for the temp buffers.

fthaler: Done.

sekelle (approved):

PR #606, opened 2026-06-18
Subset particle ID tagging from selection spheres and lists

sekelle:


ChristopherBignamini (approved):
Thanks, it looks perfect to me!

PR #614, opened 2026-07-10
Overload fix

rmcabezon:
If using double precision in all variables, max and min cuda functions fail because they don't have a dedicated double overload. I added a possible version that compiles and run.

Sorry for the dirty PR. I used the branch I was working on top of the SLR PR. The only relevant change is in the last commit. In file domain/include/cstone/traversal/ijloop/atomic_update_ptr.cuh


fthaler:
I cleaned the PR from all unrelated changes, used a non-iterative implementation based on 64bit integer atomics instead of CAS, and added unit tests.

sekelle (approved):
LGTM

PR #627, opened 2026-07-17
Simplify h-nc update

sekelle:
Update h and nc in one place.


sekelle:
Overlooked that `nc` is passed to domain as scratch

PR #628, opened 2026-07-21
Fix Non-Covergence Condition in Iterative Smoothing Length Update and Limit Neighbor Search Radius

fthaler:
- Fixes incorrect convergence check after the iterative smoothing length update.
- Sets `h[i]` to infinity in case of convergence failure.
- Skips neighbor search for particles with infinite `h[i]`.


sph/include/sph/kernels.hpp
```diff
@@ -51,7 +51,11 @@ HOST_DEVICE_FUN void updateHIterative(unsigned ng0, unsigned ngmax, const cstone
         ncSph = 1 + findNeighbors(i, x, y, z, h, treeView, box, ngmax);
     }
 
-    if (iteration == maxIteration && (ngmin > ncSph || (ncSph - 1) > ngmax)) { ncSph = 1; }
+    if (ngmin > ncSph || (ncSph - 1) > ngmax)
+    {
+        ncSph = 1;
+        h[i]  = 1e-12;
```
sekelle:
```suggestion
        h[i]  = std::numeric_limits<T>::infinity();
```

sekelle:
Can we route all loads outside of `ijLoop` on `h` through a function like `loadAtIndexIfPtr` and perform the `std::isinf` check only there and avoid infinity values ever ending up in `iData` or `jData` tuples?

fthaler:
Problem is that we sometimes want the infinite value inside iData and jData (when we pass it to the kernels) and sometimes we don’t (when checking closeness).

sekelle:
Could we use one function for loading h-data that has the `inf -> 0.0` transform whenever we want to check for overlaps? I think it would make it easier to document and less likely to miss this requirement in future changes.

fthaler:
> Could we use one function for loading h-data that has the `inf -> 0.0` transform whenever we want to check for overlaps? I think it would make it easier to document and less likely to miss this requirement in future changes.

I tried to unify the functionality as far as possible. Further, I added an infinite h value to the ijloop unit test, which checks the expected behavior at least for all neighborhoods.

sekelle:
Merged with two minor modifications:
* Reverted the test, as it didn't really increase coverage. The `inf` smoothing length was added and filtered in the test only.
* Moved h-modifying functions to `findneighbors.hpp` and also use a wrapper from `kernels.hpp` to set `h` to an invalid value. This enables us to change the signalling value if we need to.

sekelle (approved):

PR #629, opened 2026-07-22
Improvements for Clustered Neighbor List

fthaler:
- Use scratch buffer allocated in global memory instead of shared memory, allows for much larger `ncmax`.
- Use much larger `ncmax` in SPH-EXA. This might increase neighbor search times for small simulations significantly, but is barely noticeable on large-scale scenarios.


sekelle (approved):

sekelle (approved):

sekelle:
Looks good overall, but there might be a subtle bug: I don't see why the total linear and angular momentum of the isobaric cube test case would be affected by this change.

sekelle:
compare with `OMP_NUM_THREADS=64 srun  -N1 --ntasks-per-node=4 -c72 ~/cuda-vars.sh ./sphexa-cuda --init isobaric-cube --glass ~/50c.h5 -n 50 -s 10`

One a single rank, the results match the reference. So could be that only very small particle numbers are affected.

fthaler:
> compare with `OMP_NUM_THREADS=64 srun -N1 --ntasks-per-node=4 -c72 ~/cuda-vars.sh ./sphexa-cuda --init isobaric-cube --glass ~/50c.h5 -n 50 -s 10`
> 
> One a single rank, the results match the reference. So could be that only very small particle numbers are affected.

Good find, one of the changes was actually incorrect for symmetric lists. It introduced errors in rare cases which were probably caught by this test. Fixed now.

PR #631, opened 2026-07-29
Reductions in Neighborhood API

fthaler:
- Adds support for (single-rank) global reductions to the neighborhood API.
- Employs the new reductions for time step size calculation.

### DIFF
--- a/domain/include/cstone/traversal/ijloop/gpu_alwaystraverse.cuh
+++ b/domain/include/cstone/traversal/ijloop/gpu_alwaystraverse.cuh
@@ -21,6 +21,7 @@
 #include "cstone/cuda/memory.cuh"
 #include "cstone/findneighbors.hpp"
 #include "cstone/primitives/warpscan.cuh"
+#include "cstone/traversal/ijloop/atomic_update_ptr.cuh"
 #include "cstone/traversal/ijloop/common.hpp"
 #include "cstone/tree/octree.hpp"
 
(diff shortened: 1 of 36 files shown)


domain/include/cstone/traversal/ijloop/gpu_alwaystraverse.cuh
```diff
@@ -127,44 +152,71 @@ struct GpuAlwaysTraverseNeighborhood
...
-                Interaction&& interaction,
-                Postamble&& postamble,
+                Interaction const& interaction,
+                Postamble const& postamble,
+                Reduction const& reduction,
                 GroupView const& groups) const
     {
-        if (groups.numGroups == 0) return;
+        using Types                    = decltype(types(x, y, z, h, input, output, interaction, postamble, reduction));
+        using ReductionResult          = Types::ReductionResult;
```
sekelle, line 182: What if we want know what type `ReductionResult` can be? It's a pretty long chain that finally gets us to a static assert in a shuffle operation that reveals that `ReductionResult` has to be a trivially copyable type.

This is just to illustrate that the code has become very hard to read, and this change makes it harder still.  But all IJ-loop passes have quite a few constraints on the types in common, e.g. `input` and `output` tuples of field pointers. The function object `Interaction` can't just be any class, the signature of its operator() is determined by `input`, `output`, `Tc` and `ThP`, and so on.

We can probably make the ij-interface easier to understand and more self-documenting (did I mention that essential docs are largely missing?) if we put these requirements in concepts.

sekelle (changes requested):

sekelle:
I believe this breaks the block time steps. Courant time steps are not longer stored per particle, I don't see how it can work.

fthaler:
> I believe this breaks the block time steps. Courant time steps are not longer stored per particle, I don't see how it can work.

This is not true. It only drops per-particle storage of time step restrictions for the STD propagator, where no BDT variant exists. For VE, it keeps the per-particle data, but moves the global reduction into the ij-loop kernel. In the VE-BDT setting, there’s an additional group reduction kernel (as before, but it only does group-internal reductions, not anymore the global reduction).

PR #635, opened 2026-08-03
Don’t Treat Compiler Warning as Errors by Default

fthaler:
Only set `CMAKE_COMPILE_WARNING_AS_ERROR` if `SPH_EXA_WARNING_AS_ERROR` is found in the CMake environment.


sekelle (approved):
We should turn the env into Spack variant with the next CI PR.

PR #641, opened 2026-08-11
Ij interface concept

sekelle:


domain/include/cstone/traversal/ijloop/ijloop.hpp
```diff
@@ -85,33 +85,170 @@ struct max
 namespace detail
 {
 
+template<class T>
+constexpr auto unwrapModifiersImpl(T const& result)
+{
+    if constexpr (util::HasValueMember<T>)
```
fthaler, line 91: Maybe a bit too general? Could lead to surprises if a somebody supplies user-defined. However, that is slightly non-trivial anyway, so maybe not a real concern. But a specific `IsModifierType` might be safer.

domain/include/cstone/traversal/ijloop/ijloop.hpp
```diff
@@ -85,33 +85,170 @@ struct max
...
+    else
+        return result;
+}
+
+} // namespace detail
+
+template<class... Ts>
+constexpr auto unwrapModifiers(std::tuple<Ts...> const& value)
+{
+    return util::tupleMap([](auto const& v) { return detail::unwrapModifiersImpl(v); }, value);
```
fthaler, line 102: I don’t remember if `detail::unwrapModifiersImpl` is used anywhere else. If not, get rid of it and inline it here.

domain/include/cstone/traversal/ijloop/ijloop.hpp
```diff
@@ -85,33 +85,170 @@ struct max
...
+struct IsTupleOfPointers : std::false_type
+{
+};
+
+template<class... Ts>
+struct IsTupleOfPointers<std::tuple<Ts...>>
+    : std::bool_constant<(std::is_pointer_v<Ts> && ...) &&
+                         (std::is_trivially_copyable_v<std::remove_pointer_t<Ts>> && ...)>
+{
+};
```
fthaler, line 123: Personally, I find variable templates more concise and readable where they can be applied (like here), but that’s probably a matter of choice.

domain/include/cstone/traversal/ijloop/ijloop.hpp
```diff
@@ -85,33 +85,170 @@ struct max
...
+
+/*! @brief Maps a std::tuple of pointers to a std::tuple of the pointee types. */
+template<class T>
+struct DereferencedTuple;
+
+template<class... Ts>
+struct DereferencedTuple<std::tuple<Ts...>>
```
fthaler, line 130: I guess this should only accept tuples of pointers?

domain/include/cstone/traversal/ijloop/ijloop.hpp
```diff
@@ -85,33 +85,170 @@ struct max
...
+ *
+ * @tparam Tc              types of x,y,z coordinates
+ * @tparam ThP             type of h, pointer to floating_point if search radius per particle is variable
+ * @tparam Input           tuple of input particle field pointers
+ * @tparam Output          tuple of output particle field pointers
+ * @tparam Interaction     function object satisfying the PairInteraction concept
+ * @tparam Postamble       function object satisfying the ValidPostamble concept
+ */
+template<std::floating_point Tc,
+         class ThP,
```
fthaler, line 201: An additional trivial concept for `ThP` would probably also help.

domain/include/cstone/traversal/ijloop/ijloop.hpp
```diff
@@ -85,33 +85,170 @@ struct max
...
+struct IjLoopData
+{
+    //! @brief The tuple input data for a single particle in an i-j interaction,
+    using ParticleDataType = ParticleData<Tc, ThP, Input>;
+
+    //! @brief Type of search radii, e.g. smoothing lengths
+    using RadiusType = std::remove_pointer_t<ThP>;
+
+    //! @brief What an i-j interaction returns
+    using Result = decltype(std::declval<Interaction>()(std::declval<ParticleDataType>(),
```
fthaler, line 215: Maybe `InteractionResultType`? The name `Result` is a relict from times where no postambles existed, but when touching it now, we should probably update it.

domain/include/cstone/traversal/ijloop/ijloop.hpp
```diff
@@ -85,33 +85,170 @@ struct max
...
+
+    //! @brief the i-j interaction kernel
+    Interaction interaction;
+    //! @brief Post-processing to apply to the Result after the j-loop
+    Postamble postamble;
+};
+
+//! @brief Convenience factory to construct an @p IjLoopData with explicit Tc and ThP and deduced tuple/functor types.
+template<class Tc, class ThP, class Input, class Output, class Interaction, class Postamble>
+auto makeIjLoopData(Input&& input, Output&& output, Interaction&& interaction, Postamble&& postamble)
```
fthaler, line 234: I wonder if deduction guides could solve this? (Maybe not, no time to think more deeply.)
Also perfect forwarding is probably an overkill, const references should be enough and simplify the code a bit.

fthaler (changes requested):
Looks good overall, only some minor issues found.

domain/include/cstone/traversal/ijloop/ijloop.hpp
sekelle: done

domain/include/cstone/traversal/ijloop/ijloop.hpp
sekelle: done

domain/include/cstone/traversal/ijloop/ijloop.hpp
sekelle: It could be inlined with a self-referencing lambda, but would be much more unreadable than the separate helper.

domain/include/cstone/traversal/ijloop/ijloop.hpp
sekelle: I don't think the concern is really that big. If you add a new type of symmetry or reduction tag, ensuring correct ij-loop behavior will be the bigger issue than the unwrapping.

domain/include/cstone/traversal/ijloop/ijloop.hpp
sekelle: done

domain/include/cstone/traversal/ijloop/ijloop.hpp
fthaler: My bad, I missed the recursion.

fthaler (approved):

domain/include/cstone/traversal/ijloop/ijloop.hpp
sekelle: done

PR #642, opened 2026-08-12
Experimental: Support Direct Kernel Evaluation 

fthaler:
Implements direct numerical evaluation of smoothing kernels instead of using LUTs. On MI300A, this speeds up some kernels by about 25%, on GH200 seems to be 5–10%.


domain/include/cstone/primitives/fastmath.hpp
sekelle, line 1: This file can go to the `primitives` folder, namespace `cstone`, where the other elementary numerical operations are living.

domain/include/cstone/util/fastmath.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+#pragma once
+
+#include <cmath>
+
+namespace util::fastmath
+{
+
+[[gnu::optimize("-ffast-math")]]
+constexpr float sin(float x)
```
sekelle:
```suggestion
#define CSTONE_FAST_MATH [[gnu::optimize("-ffast-math")]]

CSTONE_FAST_MATH constexpr float sin(float x)
```

sekelle (changes requested):

domain/include/cstone/primitives/fastmath.hpp
fthaler: Done.

domain/include/cstone/util/fastmath.hpp
fthaler: Done.

sekelle (approved):

